PR #3, opened 2021-02-08
Add handling for if the user ends the session by timing out

rudigiesler:
If the user times out in the Turn conversation claim, the transport will send us a session close message. We should reply to this message letting the user know that they've timed out, and how to get back in


OMosimege (approved):
👍

PR #4, opened 2021-03-12
Add metrics and healthchecks

rudigiesler:
This adds some basic metrics, as well as healthchecks for amqp and redis


erikh360 (approved):
Nice!

PR #6, opened 2021-04-09
Implement proper user flow

rudigiesler:
This replaces the demo flow with the actual user flow for vaccine eligibility

### DIFF
--- a/vaccine/application.py
+++ b/vaccine/application.py
@@ -16,33 +16,45 @@ class Application(BaseApplication):
         return await self.go_to_state("state_occupation")
 
     async def state_occupation(self):
+        not_sure = "\n".join(
+            [
+                "*Health Care Workers* include doctors, nurses, dentists, pharmacists, "
+                "medical specialists and all people involved in providing health "
+                "services such as cleaning, security, medical waste disposal and "
+                "administrative work.",
+                "",
+                "*Essential Workers* include police officers, miners, teachers, people "
+                "working in security, retail, food, funeral, banking and essential "
+                "muncipal and home affairs, border control and port health services.",
+            ]
+        )
+
+        async def next_state(choice: Choice):
+            if choice.value == "not_sure":
+                self.send_message(not_sure)
+                return "state_occupation"
+            return "state_age"
+
         return ChoiceState(
             self,
             question="\n".join(
                 [
-                    "Welcome to the vaccine eligibility service.",
-                    "Please answer a few questions so that we can determine your "
-                    "eligibility.",
+                    "◼️◻️◻️◻️◻️",
                     "",
-                    "What is your current occupation?",
+                    "Which of these positions or job titles describes your current "
+                    "employment:",
+                    "",
                 ]
             ),
             choices=[
-                Choice("unemployed", "Unemployed"),
-                Choice("retired", "Retired"),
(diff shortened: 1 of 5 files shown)


vaccine/application.py
```diff
@@ -43,30 +118,213 @@ async def check_age(content: Optional[str]):
...
 
-    async def state_end(self):
+    async def state_location(self):
+        return FreeText(
+            self,
+            question="\n".join(
+                [
+                    "◼️◼️◼️◼️◻️",
+                    "",
+                    "Where do you live? (Reply with name of suburb, town, settlement "
```
miltontony, line 139: Will we not need to use Maps API to look up the location like we do for Health Check?

vaccine/application.py
```diff
@@ -43,30 +118,213 @@ async def check_age(content: Optional[str]):
...
+                ]
+            ),
+            next="state_comorbidities",
+        )
+
+    async def state_comorbidities(self):
+        not_sure = "\n".join(
+            [
+                "Having one or more specific chronic conditions could impact which "
+                "phase you are in. These conditions include:",
```
miltontony, line 150: Do you know whether this is the full list? Just thinking to confirm with Neville

DevChima:
Looks good

vaccine/application.py
```diff
@@ -1,40 +1,115 @@
...
+                        [
+                            "We're sorry, but you've taken too long to reply and your "
+                            "session has expired.",
+                            "If you would like to continue, you can at anytime by "
+                            "typing the word *VACCINE*.",
+                            "",
+                            "Reply *MENU* to return to the main menu",
+                        ]
+                    ),
+                    continue_session=False,
```
miltontony, line 26: Does the framework allow for continuation?

miltontony (approved):
Looks good, just a few questions

vaccine/application.py
rudigiesler: Yeah we can, will check with Neville because it didn't mention this in the sheet (it also doesn't mention location pins)

vaccine/application.py
rudigiesler: This content is from Neville, but will confirm

vaccine/application.py
rudigiesler: Yup, by default it continues session, this is to say that we want to end the session on our side (although it should already be ended because we're getting a message from the user that they've ended the session)

vaccine/application.py
Neville-Tietz: These are the current co-morbidities approved by NDOH

vaccine/application.py
Neville-Tietz: Have amended the copy in the source spreadsheet as follows:

Neville-Tietz: ◼️◼️◼️◼️◻️

Please tell us where you live.

You can share your location 📍 using WhatsApp by following these steps:
- Tap the attach icon (+ or  📎) on the bottom of WhatsApp. 
- Select location
- This will bring up a map with a pin showing your current location. Tap *Send Your Current Location* if this is your address.
- You can also choose from various other nearby locations if you think that the GPS is a bit off.

Or TYPE the name of your Suburb, Township, Town or Village (or nearest)?

miltontony (approved):
👍

PR #7, opened 2021-04-13
Vaccine registration USSD flow

rudigiesler:


vaccine/vaccine_reg_ussd.py
```diff
@@ -0,0 +1,363 @@
...
+                    "coronavirus!",
+                    "",
+                    "Are you 40 years or older?",
+                ]
+            ),
+            choices=[
+                Choice("state_identification_type", "Yes"),
+                Choice("state_under40_notification", "No"),
+            ],
+            error="Self registration is currently only available to those 40 years or "
```
miltontony, line 43: What would trigger this error? invalid input?

vaccine/vaccine_reg_ussd.py
```diff
@@ -0,0 +1,363 @@
...
+            "older. Please tell us if you are 40 years of age or older?",
+        )
+
+    async def state_under40_notification(self):
+        return ChoiceState(
+            self,
+            question="Self registration is only available to those 40 years or older. "
+            "Can we notifify you by SMS on this number when this changes?",
```
miltontony, line 51: typo on `notifify`, but not an issue since we need to update this with the official copy when it's ready

vaccine/vaccine_reg_ussd.py
```diff
@@ -0,0 +1,363 @@
...
+        id_type = self.ID_TYPES[self.user.answers["state_identification_type"]].value
+        id_number = self.user.answers["state_identification_number"][:20]
+        return MenuState(
+            self,
+            question="\n".join(
+                [
+                    "Confirm the following:",
+                    "",
+                    f"{first_name} {surname}",
+                    id_type,
```
miltontony, line 187: We were not going to include the id_type in the confirmation question. Also, how many characters do we have to play around in this screen? Makes sense to remove this and only confirm Firstname, Surname, ID

vaccine/vaccine_reg_ussd.py
```diff
@@ -0,0 +1,363 @@
...
+                    id_number,
+                ]
+            ),
+        )
+
+    async def state_province(self):
+        return ChoiceState(
```
miltontony, line 207: Add `#TODO` for using the Province selector using the EVDS uuids for Province. Alternatively map them in the end

miltontony (commented):
First pass review - looks good so far

vaccine/vaccine_reg_ussd.py
rudigiesler: Yup. So the `MenuState` is very similar to the `ChoiceState`. By default, for `1. Yes`, you can either respond with `1` or `yes`. If you respond with anything else, you'll get the error message, with the list of options.

vaccine/vaccine_reg_ussd.py
rudigiesler: So I've limited first name and last name to 27 characters, and identification to 20 characters. For the longest id type, on the error message (that's longer than the question message), that gives us 158 characters if all those fields are at their maximum, to 2 characters to spare. But we can adjust those maximums.

vaccine/vaccine_reg_ussd.py
rudigiesler: Looking at the API, it seems like all they want is the lowest tier result (ie. suburb), and then they infer the parent fields from that. But I'll add a todo here, since we'll need the province ID to filter the suburbs.

vaccine/vaccine_reg_ussd.py
miltontony: Ok, but I think we should remove `id_type` and increase the limits accordingly.

miltontony: @Neville-Tietz Please review this suggestion - it has copy implications

Neville-Tietz: @miltontony @rudigiesler Can we drop id_type but include a shortened type as in "ID#:identification_number"

miltontony (approved):

miltontony (approved):

PR #23, opened 2021-04-16
Move terms for ussd to top

miltontony:


vaccine/tests/test_vaccine_reg_ussd.py
```diff
@@ -988,19 +988,21 @@ async def test_medical_aid_invalid():
 @pytest.mark.asyncio
 async def test_terms_and_conditions():
     u = User(
-        addr="27820001001", state=StateData(name="state_medical_aid"), session_id=1
+        addr="27820001001",
+        state=StateData(name="state_terms_and_conditions_3"),
```
rudigiesler, line 992: This should be the state before `state_terms_and_conditions`, so that when you answer that state, it displays `state_terms_and_conditions`. In this case, it's `state_age_gate`

vaccine/tests/test_vaccine_reg_ussd.py
```diff
@@ -1105,9 +1107,10 @@ async def test_terms_and_conditions_3_invalid():
 
 @pytest.mark.asyncio
 async def test_state_success(evds_mock):
+    evds_mock.app.errormax = 1
```
rudigiesler, line 1110: I'm not sure what this is for? I think we've got a separate test to test this

vaccine/tests/test_vaccine_reg_ussd.py
```diff
@@ -1139,10 +1142,10 @@ async def test_state_success(evds_mock):
         "information and appointment details will be sent via SMS."
     )
     assert reply.session_event == Message.SESSION_EVENT.CLOSE
-    assert u.state.name == "state_age_gate"
```
rudigiesler, line 1142: We should still end up in `state_age_gate` after we've submitted to EVDS

vaccine/tests/test_vaccine_reg_ussd.py
miltontony: Yeah, it was a copy/paste error. fixing

vaccine/tests/test_vaccine_reg_ussd.py
miltontony: yup, that's correct. I thought it was redundant since we actually have a session close event. I've re-added this assert

rudigiesler (approved):

PR #26, opened 2021-04-19
Add more ussd healthcheck tests

erikh360:


vaccine/healthcheck_ussd.py
```diff
@@ -768,6 +771,8 @@ async def state_tracing(self):
         )
 
     async def state_submit_data(self):
+        if self.user.answers.get("state_tracing") == "restart":
```
erikh360, line 774: I'm not sure if there is a better way to deal with this.

rudigiesler: Yeah I think I've done this for the vaccine registration as well. Perhaps a better way is, for the jssandbox, you can have `next` be a function or a string, that receives the `choice` that the user made, and returns a string of which state to go to next. But happy with this for now, until we implement that.

rudigiesler (approved):

PR #29, opened 2021-04-19
add last of healthcheck ussd tests

erikh360:


vaccine/tests/test_healthcheck_ussd.py
```diff
@@ -1007,6 +1194,191 @@ async def test_state_breathing():
     )
 
 
+pytest.mark.asyncio
```
rudigiesler, line 1197: I think this should be an `@pytest.mark.asyncio`

rudigiesler (approved):

PR #36, opened 2021-04-19
add ability to search for medical aid scheme

miltontony:


rudigiesler (approved):
We could possibly cache the `{v.strip().lower(): k}` and `{k: v}` dictionaries, otherwise I think we'll be recreating those dictionaries N times for every search, but it's a small list so probably not that important

rudigiesler (approved):

PR #52, opened 2021-04-20
added country search ussd

miltontony:


erikh360 (approved):
I think its possible for the country list message to go over 160 characters, but :+1:

PR #56, opened 2021-04-21
Update cached api data 21 4 2021

miltontony:


vaccine/data/medscheme.py
```diff
@@ -15,21 +15,14 @@ def data(self):
...
-        schemes_kv = {k: v for k, v in schemes}
-        return lowercase, schemes_kv
+        return {i["value"]: i["text"] for i in self.data}
 
     def search_for_scheme(self, search_text):
-        search_text = search_text.strip().lower()
-        lowercase, _ = self.schemes
-        possibilities = difflib.get_close_matches(search_text, lowercase.keys(), n=3)
-        return [(lowercase[p], self.scheme_name(lowercase[p])) for p in possibilities]
+        possibilities = process.extract(search_text, self.data, limit=3)
```
rudigiesler, line 21: Interesting that we can use `self.data` here (a list of dictionaries), I would've expected us to have to use `self.schemes` (a dictionary of IDs and values to search)

vaccine/tests/test_vaccine_reg_ussd.py
```diff
@@ -841,7 +841,7 @@ async def test_suburb_search():
     )
     app = Application(u)
     msg = Message(
-        content="western cape",
+        content="9",
```
rudigiesler, line 844: We should still be able to use the label `western cape` here, but with alphabetical sorting, western cape should always be last, so this should be fine

rudigiesler (approved):

vaccine/tests/test_vaccine_reg_ussd.py
miltontony: I wanted to ensure that the tests reflect the user choice more accurately so that the `value` doesn't slip through rather than the choice.

vaccine/data/medscheme.py
miltontony: Very true - should I change this?

PR #62, opened 2021-04-23
Suburb search improvements

rudigiesler:
This PR uses `extractBests` to clean up the code a bit, as well as returns the municipalities in order so that the lowest score municipalities get cut off


miltontony (approved):
👍

PR #65, opened 2021-04-23
Add other option for suburb search

miltontony:


vaccine/vaccine_reg_ussd.py
```diff
@@ -406,16 +406,23 @@ async def state_municipality(self):
...
+        async def next_state(choice: Choice):
+            if choice.value == "other":
+                return "state_suburb_search"
+            return "state_suburb"
+
         question = "Please select your municipality"
+        choices = [Choice(k, v[:30]) for k, v in results]
         choices = enforce_character_limit_in_choices(
-            [Choice(k, v[:30]) for k, v in results], 160 - len(question)
+            choices, 160 - len(question) - len("X. Other ")
```
rudigiesler, line 417: I think we'll need to take into account 2 newlines here, one after the question, and one between the choices and the other choice

vaccine/data/suburbs.py
```diff
@@ -110,7 +110,7 @@ async def ussd_search(self, province_id, search_text, municipality_id=None):
             ]
             return (True, self._filter_duplicate_municipalities(municipalities))
         else:
-            return (False, [(id, name) for name, _, id in possibilities[:3]])
+            return (False, [(id, name) for name, _, id in possibilities])
```
rudigiesler, line 113: I think we'll need to do this now in the state, as the logic is still `if len(possibilities) > 3` in here, where maybe we should do `if len(possibilities) > len(possibilities_that_fit)` in the state. But also happy just to keep it at 3, then you can ignore this comment.

vaccine/vaccine_reg_ussd.py
miltontony: I added an extra space after the "Other" option for the first linebreak.. I'll add another for the linebreak after the question

vaccine/data/suburbs.py
miltontony: Happy to leave it at 3 for now 😃

rudigiesler (approved):

PR #67, opened 2021-04-23
Don't submit residential location

miltontony:


rudigiesler (approved):
We'll also need to update the API version number for this. Not sure if we want to do that in this PR, or separately.

PR #70, opened 2021-04-25
Add ussd dialing shortcut with id number

miltontony:


vaccine/vaccine_reg_ussd.py
```diff
@@ -136,12 +134,33 @@ async def state_terms_and_conditions_2(self):
...
+            elif (
+                self.user.answers.get("state_identification_type")
+                and self.user.answers.get("state_identification_number")
+                and "state_dob_year" not in self.user.answers
+                and self.user.answers.get("state_dob_month")
+                and self.user.answers.get("state_dob_day")
+                and self.user.answers.get("state_gender")
+            ):
+                return "state_dob_year"
+            return "state_identification_type"
```
rudigiesler, line 156: Instead of having this large if statement (which might break if the order of the states changes), what if we just have inside `state_identification_type`
```
if self.user.answers.get("state_identification_type"):
    return await self.go_to_state("state_identification_number")
```
And in `state_identification_number` we have
```
if self.user.answers.get("state_identification_number"):
    return await self.go_to_state(next_state)
```

That way, it will just skip through the `identification_type` and `identification_number` states if we already have a value for them, and reuse the "skip DoB if we already have it" logic that is already there.

rudigiesler (commented):
Just one comment around a way we could possibly simplify things, looks good otherwise

rudigiesler (approved):

PR #72, opened 2021-04-26
Add loadtest for vaccine registration USSD

rudigiesler:


vaccine/tests/vaccine_reg_loadtest.py
```diff
@@ -0,0 +1,111 @@
+import gzip
+import json
+from datetime import date
+
+from locust import HttpUser, between, task
+
+from vaccine.tests.generate_test_ussd_data import generate_vaccine_registration
+
+PATH = "/aat/ussd"
+ADDRESS = "*120*4216*2#"
```
miltontony, line 10: I think these should be ENV variables so that they are not in a public repo

miltontony (approved):

PR #74, opened 2021-04-28
Add throttling config

rudigiesler:
Adds a throttling config value to limit the amount of people that go through to complete the registration. This is there in case upstream systems are struggling with the traffic, then we can switch it on temporarily to reduce traffic


erikh360 (approved):

vaccine/vaccine_reg_whatsapp.py
```diff
@@ -101,6 +102,9 @@ async def state_timeout(self):
     async def state_age_gate(self):
         self.user.answers = {}
 
+        if random.random() <= config.THROTTLE_PERCENTAGE * 100:
```
miltontony, line 105: I don't think this is correct - `random.random()` always returns a value less than `1`

rudigiesler: Whoops, thanks

miltontony (approved):

PR #94, opened 2021-05-20
Improved application testing

rudigiesler:
This adds an AppTester helper class, that reduces boilerplate when testing applications


miltontony (approved):
Looks good to me

PR #95, opened 2021-05-24
USSD privacy policy changes

erikh360:


vaccine/healthcheck_ussd.py
```diff
@@ -155,6 +166,10 @@ async def state_start(self):
         self.save_answer(
             "state_preexisting_conditions", data["data"].get("preexisting_condition")
         )
+        self.save_answer(
+            "state_privacy_policy_accepted",
+            data["data"].get("privacy_policy_accepted", False),
```
rudigiesler, line 171: For flow-results, the spec doesn't have booleans, so we usually store these as yes/no strings.

vaccine/healthcheck_ussd.py
```diff
@@ -260,12 +270,60 @@ async def state_terms(self):
...
+                            json=data,
+                        )
+                        response.raise_for_status()
+                        break
+                    except HTTP_EXCEPTIONS as e:
+                        if i == 2:
+                            logger.exception(e)
+                            return await self.go_to_state("state_error")
+                        else:
+                            continue
```
rudigiesler, line 310: `state_privacy_policy` will be called twice, once to give the user the choices, and again when we receive the response to validate the user's choice, which would cause 2 smses to be sent.

We can avoid this by having a middle state, that sends the SMS, and then goes to this state without asking anything from the user.

erikh360: ah good catch thanks

rudigiesler (approved):

PR #100, opened 2021-06-02
Add sentry context for answers api

erikh360:


vaccine/worker.py
```diff
@@ -185,31 +186,36 @@ async def _push_results(self):
...
-                                for a in answers
-                            ]
-                        },
-                    }
-                },
+                json=data,
+            )
+            response_data = await response.json()
+            sentry_sdk.set_context(
+                "answer_api", {"request_data": data, "response_data": response_data}
```
rudigiesler, line 218: I think we'll need to set `raise_for_status=False` where `self.session` is defined, otherwise it will raise the exception before we get to this line. And then manually call to raise for status after we set this context.

rudigiesler (approved):

PR #106, opened 2021-06-03
Answer worker one-by-one error handling

rudigiesler:
Currently, we try to submit the whole batch, and if that fails, then we log and retry later.

For 400 errors though, it doesn't help to retry. The issue is we don't know which answer resulted in the 400, and we don't want to throw all the answers away.

This PR changes the functionality so that if there's a 400, it goes through the answers one-by-one, and when if finds a 400, it logs the error and throws that answer away.

### DIFF
--- a/vaccine/tests/test_worker.py
+++ b/vaccine/tests/test_worker.py
@@ -15,6 +15,35 @@ from vaccine.models import Answer, Event, Message, StateData, User
 from vaccine.worker import AnswerWorker, Worker, config, logger
 
 
+@pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


vaccine/tests/test_worker.py
```diff
@@ -15,6 +15,35 @@
 from vaccine.worker import AnswerWorker, Worker, config, logger
 
 
+@pytest.fixture(autouse=True)
+async def cleanup():
```
erikh360, line 19: Thanks!

erikh360 (approved):

PR #107, opened 2021-06-04
Add tb prompts at the end of the healthcheck

erikh360:


vaccine/healthcheck_ussd.py
```diff
@@ -892,6 +892,61 @@ async def state_display_risk(self):
...
+        risk = self.calculate_risk()
+
+        text = ""
+        if risk == "moderate":
+            if answers.get("state_cough") == "yes":
+                text = self._(
+                    "A cough may also be a sign of TB – a dangerous but treatable "
+                    "disease."
+                )
+            elif risk == "moderate" and answers.get("state_fever") == "yes":
```
rudigiesler, line 917: We already know that `risk == "moderate"` due to the parent if statement

rudigiesler (commented):
Looks good, just a small comment on how to simplify the logic a little bit, but looks good otherwise.

vaccine/healthcheck_ussd.py
erikh360: ah yes, the parent statement came later and I forgot to remove it

PR #109, opened 2021-06-07
Add Ask a Question flow

rudigiesler:
Adds the user for for the ask a question feature, which allows the user to ask a freetext question and get a response from our FAQs

﻿- Add ask a question flow
- Add footer option for choice state


vaccine/ask_a_question.py
```diff
@@ -0,0 +1,237 @@
...
+        data = {
+            "inbound_id": model_response["inbound_id"],
+            "feedback_secret_key": model_response["feedback_secret_key"],
+            "feedback": self.user.answers["state_display_selected_choice"],
+        }
+        async with model as session:
+            for i in range(3):
+                try:
+                    response = await session.post(
+                        url=urljoin(config.MODEL_API_URL, "/inbound/feedback"),
```
erikh360, line 194: This is cool, do they use it for training?

erikh360 (approved):
Looks good

vaccine/ask_a_question.py
rudigiesler: Yeah, but it's not for automated training. This just stores it in the database so that we can pull it later, and use it to improve the model.

PR #110, opened 2021-06-08
AAQ send question choice with feedback

rudigiesler:
Along with sending the user feedback, we should also send the question choice, so that we know which question the user selected from the top 3﻿


erikh360 (approved):
when tests are green

PR #112, opened 2021-06-09
Ask a Question UX improvements

rudigiesler:
- Copy change for question
The copy for the question state was changed to better reference the ASK keyword

- end state UX improvement
The previous UX was confusing, so instead it's been replaced with just a normal EndState, with references to the ASK keyword to start the flow again


miltontony (approved):
👍

PR #122, opened 2021-07-19
Load Zulu translations

rudigiesler:
- Add language choice screen
- Add Zulu translations
- Include compiled translations


miltontony (approved):
👍

PR #128, opened 2021-07-21
Translation fixes

miltontony:


erikh360 (approved):
looks good, not 100% sure about skip in afr but I assume its intentionally done that way

miltontony:
Yes it is. We're using SKIP as a keyword for all languages.

PR #142, opened 2021-09-15
Hotline Callback

rudigiesler:
Adds an application that has the functionality to request a callback from the helpdesk if you're having difficulty with the automated bot

- Add hotline callback first menu
- Add full name state
- Add hotline select number state
- Add hotline enter number state
- Add submission to callback API
- Get message list

### DIFF
--- a/vaccine/hotline_callback.py
+++ b/vaccine/hotline_callback.py
@@ -5,7 +5,7 @@ from vaccine.base_application import BaseApplication
 from vaccine.models import Message
 from vaccine.states import Choice, EndState, FreeText, WhatsAppButtonState
 from vaccine.utils import display_phonenumber
-from vaccine.validators import nonempty_validator
+from vaccine.validators import nonempty_validator, phone_number_validator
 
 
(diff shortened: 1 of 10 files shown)


vaccine/hotline_callback.py
```diff
@@ -0,0 +1,272 @@
...
+    HTTP_EXCEPTIONS,
+    TZ_SAST,
+    display_phonenumber,
+    get_today,
+    normalise_phonenumber,
+)
+from vaccine.validators import nonempty_validator, phone_number_validator
+
+logger = logging.getLogger(__name__)
+za_holidays = holidays.SouthAfrica()
```
erikh360, line 25: This is cool

erikh360 (approved):
looks good

vaccine/hotline_callback.py
rudigiesler: Yeah, I started with a list of datetimes for the next few years, but found this library, which handles Easter properly and everything.

PR #148, opened 2021-10-13
Vaccine certificate decoder example application

rudigiesler:
An example whatsapp application to be able to extract and display the data contained inside a vaccine certificate. Does not do any validation of the signature, so does not check for fraudulent certificates.


erikh360 (approved):
🍰

PR #162, opened 2021-12-02
Cases province latest alongside totals

rudigiesler:
This adds the number of cases in the last day in brackets next to the totals per province.


erikh360 (approved):
Nice!

PR #163, opened 2021-12-02
Add latest daily cases image

rudigiesler:
This scrapes the page for the latest posted image, and sends it along with the message

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ holidays = "^0.11.2"
 email-validator = "^1.1.3"
 zbar-py = "^1.0.4"
 opencv-python-headless = "^4.5.3"
+beautifulsoup4 = "^4.10.0"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -24,6 +24,7 @@ holidays = "^0.11.2"
 email-validator = "^1.1.3"
 zbar-py = "^1.0.4"
 opencv-python-headless = "^4.5.3"
+beautifulsoup4 = "^4.10.0"
```
erikh360, line 27: Yoh, haven't seen this in a while

erikh360 (approved):

pyproject.toml
rudigiesler: Yeah unfortunately we have to scrape, and this is the best library I know for scraping :)

PR #167, opened 2021-12-23
Handle LockNotOwnedError

rudigiesler:
There's nothing much we can do if we finish processing the message, and then we find out that we no longer own the lock. So log the exception and carry on processing.

This seems to cause an issue and crash the worker if we don't handle the exception, so we handle it. We also probably don't want to reprocess the message for this error﻿


jerith (approved):
:+1:

PR #168, opened 2021-12-23
Add timing debug logging for message processing

rudigiesler:
We keep on getting LockNotOwnedErrors, so we need to see what is taking so long that the locks are timing out


jerith (approved):
:+1:  although it might be useful to log a message id of some kind in case we have overlapping messages.

jerith (approved):
:+1:

PR #170, opened 2022-01-18
Real411 V2 API

rudigiesler:
We need to switch to the new version of the real411 API. It has a number of changes:
- Form data is now split out into separate endpoints
- Filtering of form data can now be done server side
- The identifier is now automatically sourced from the API key
- The complaint URL is now available when submitting the form
- References were switched from integers to strings

### DIFF
--- a/vaccine/real411.py
+++ b/vaccine/real411.py
@@ -1,5 +1,6 @@
 import json
 import re
+from asyncio import gather
 from base64 import b64decode
 from typing import List, Optional, Tuple
 from urllib.parse import urljoin
@@ -22,6 +23,12 @@ BLANK_PNG = b64decode(
 )
 
 
+class Real411APIException(Exception):
+    """
+    Error when interacting with the Real411 API
+    """
+
+
 def get_real411_api_client() -> aiohttp.ClientSession:
     return CachedSession(
         cache=cache_backend,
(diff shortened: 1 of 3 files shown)


vaccine/real411.py
```diff
@@ -409,7 +426,7 @@ async def state_submit_report(self):
         if not files:
             files.append({"name": "placeholder", "type": "image/png"})
 
-        form_reference, file_urls = await submit_real411_form(
+        form_reference, file_urls, self.backlink = await submit_real411_form(
```
rudigiesler, line 429: `backlink` will go into the message that we send to the user, but we don't have the copy for that yet, so it will go in a separate PR.

vaccine/tests/test_real411.py
```diff
@@ -18,36 +20,79 @@ async def real411_mock(sanic_client):
...
-    @app.route("/submit/v2", methods=["POST"])
+    @app.route("/complaint", methods=["POST"])
     def submit(request):
         app.requests.append(request)
         file_count = len(request.json["file_names"])
         return response.json(
             {
-                "complaint_ref": 1,
+                "success": True,
+                "complaint_ref": "WDM88J4P",
```
rudigiesler, line 81: References were switched to strings, just want to make sure that the tests assert that, in case we assume integer anywhere

erikh360 (approved):

PR #181, opened 2022-03-17
First two questions for ussd

Buhle79:


mqr/baseline_ussd.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class Application(BaseApplication):
+    START_SURVEY = "survey_start"
+
+    async def survey_start(self):
+        if config.RAPIDPRO_URL and config.RAPIDPRO_TOKEN:
+            msisdn = normalise_phonenumber(self.inbound.from_addr)
+            if msisdn.startswith("+"):
+                urn = ' f"whatsapp:{msisdn.lstrip(' + ')}"'
```
erikh360, line 35: I think we have a few extra single quotes and spaces here: 
urn = f"whatsapp:{msisdn.lstrip('+')}"

erikh360: The if statement is also not necessary, the lstrip will just not do anything if + isn't there

mqr/baseline_ussd.py
```diff
@@ -0,0 +1,119 @@
...
+                urn = msisdn
+
+            for i in range(3):
+                try:
+                    contact = rapidpro.get_contacts(urn=urn).first(
+                        retry_on_rate_exceed=True
+                    )
+
+                    if not contact:
+                        self.save_answer("returning_user", "no")
```
erikh360, line 46: Don't think we need `returning_user`, just check that they accepted mqr_consent and have a arm assigned

vaccine/base_application.py
```diff
@@ -80,7 +80,8 @@ async def process_message(self, message: Message) -> List[Message]:
                 await state.display(message)
             else:
                 await state.process_message(message)
-        except Exception:
+        except Exception as e:
+            print('*******', e)
```
erikh360, line 84: can remove these

mqr/utils.py
```diff
@@ -0,0 +1,7 @@
+from temba_client.v2 import TembaClient
+
+import vaccine.healthcheck_config as config
+
+rapidpro = None
+if config.EXTERNAL_REGISTRATIONS_V2:
```
erikh360, line 6: Don't need these lines and EXTERNAL_REGISTRATIONS_V2

vaccine/healthcheck_config.py
```diff
@@ -10,3 +10,4 @@
 RAPIDPRO_TOKEN = environ.get("RAPIDPRO_TOKEN")
 RAPIDPRO_PRIVACY_POLICY_SMS_FLOW = environ.get("RAPIDPRO_PRIVACY_POLICY_SMS_FLOW")
 TB_USSD_CODE = environ.get("TB_USSD_CODE", "")
+EXTERNAL_REGISTRATIONS_V2 = environ.get("EXTERNAL_REGISTRATIONS_V2")
```
erikh360, line 13: Don't need this

erikh360 (commented):
few small comments

mqr/baseline_ussd.py
Buhle79: If we don't have them, should it end the survey? @erikh360

mqr/baseline_ussd.py
```diff
@@ -0,0 +1,116 @@
...
+            msisdn = normalise_phonenumber(self.inbound.from_addr)
+            urn = ' f"whatsapp:{msisdn.lstrip(' + ')}"'
+
+            for i in range(3):
+                try:
+                    contact = rapidpro.get_contacts(urn=urn).first(
+                        retry_on_rate_exceed=True
+                    )
+
+                    if not contact:
```
erikh360, line 42: Happy for this to be sorted in a next PR but this says if the rapidpro contact doesn't exist continue to survey start. We want the contact to exist and we want it to be an MQR contact

mqr/utils.py
```diff
@@ -0,0 +1,6 @@
+from temba_client.v2 import TembaClient
+
+import vaccine.healthcheck_config as config
+
+
+rapidpro = TembaClient(config.RAPIDPRO_URL, config.RAPIDPRO_TOKEN)
```
erikh360, line 6: Doesn't look like we are actually using any of this, can remove this file.

erikh360 (commented):
few more small ones

mqr/utils.py
Buhle79: I am using it to get rapidpro contact

erikh360: I see, then we probably don't need the `def get_rapidpro():` function. 

Just not sure if `TembaClient` will work in async. But lets see

erikh360: You can then also move the:
`if config.RAPIDPRO_URL and config.RAPIDPRO_TOKEN:`
to here.

and replace it with `if rapidpro` where you are using it.

erikh360 (commented):
added some comments

mqr/baseline_ussd.py
```diff
@@ -0,0 +1,506 @@
...
+from vaccine.utils import HTTP_EXCEPTIONS, normalise_phonenumber
+
+logger = logging.getLogger(__name__)
+
+
+class Application(BaseApplication):
+    START_SURVEY = "survey_start"
+
+    async def survey_start(self):
+        if config.RAPIDPRO_URL and config.RAPIDPRO_TOKEN:
```
erikh360, line 16: We don't need this if statement

mqr/baseline_ussd.py
```diff
@@ -0,0 +1,506 @@
...
+            msisdn = normalise_phonenumber(self.inbound.from_addr)
+            urn = f"whatsapp:{msisdn.lstrip(' + ')}"
+
+            if rapidpro:
+                for i in range(3):
+                    try:
+                        contact = rapidpro.get_contacts(urn=urn).first(
+                            retry_on_rate_exceed=True
+                        )
```
rudigiesler, line 25: It's not a good idea to have blocking code running inside the event loop, it's explained well here, even though it's describing for nodejs, it's similar concepts in python: https://nodejs.org/en/docs/guides/dont-block-the-event-loop/#don-t-block-the-event-loop

The easiest way to solve this is to use `run_in_executor`, which runs the sync function in a threadpool: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor , or by using an async HTTP client.

erikh360: Thanks @rudigiesler , I thought so.

@Buhle79 please ignore this comment for this PR, I'll fix in the next PR. Just going to use the async http client.

erikh360 (approved):
Go ahead and merge, I'll fix things in the next PR

PR #200, opened 2022-05-18
Add allow skip to ChoiceState

erikh360:


rudigiesler (approved):
Looks good to me.

The only issue I see with this is is that the button value is static (can't set it in the flow), which makes reusing it in the future a bit difficult (which might not be an issue, but as it stands you can't customise one state to be different to another), and no translations.

A more flexible way could be to add a `buttons` argument instead, that takes in a list of `Choice`s. Then in the display it adds those to `helper_metadata["buttons"]`, and in the `_get_choice` it checks the content across both the `self.choices` and `self.buttons`. ie. adding buttons is just like adding more choices to the list, but they get displayed differently.

But I'm happy with the way it is now, it works for what we need it and is simple

erikh360:
I like that idea! I'm going to merge for now and if we need it we can implement it that way

PR #201, opened 2022-05-18
Buttons on FreeText and ChoiceState

erikh360:


rudigiesler (approved):
:+1: looks good

Month day year is an odd order, and makes it difficult to validate day. The validation won't catch all invalid dates (eg 30 february), but you can't actually validate until you have the year as well (due to leap years), so by asking year least you have this weird flow where if you want to properly validate the day, you have to tell the user an error after they type in the year. Just something to think about if you want to have better dob validation.

erikh360:
Thanks, I'll raise with Chuk when he is back.

PR #207, opened 2022-05-24
Add sectioned choicestate for main menu

erikh360:


vaccine/states.py
```diff
@@ -254,3 +254,32 @@ async def display(self, message: Message):
...
+        self.sections = sections
+        self.separator = separator
+
+    @property
+    def _display_choices(self) -> str:
+        lines = []
+
+        i = 1
+        for section, choices in self.sections.items():
+            lines.append(f"*{section}*")
```
rudigiesler, line 277: Putting formatting here seems odd (especially since this is whatsapp specific). Surely we could put that in when we create the state in `yal/mainmenu.py`, rather than having it hardcoded here. eg. `"*CHAT SETTINGS*"` instead of `"CHAT_SETTINGS"`

vaccine/states.py
```diff
@@ -254,3 +254,32 @@ async def display(self, message: Message):
                 ],
             },
         )
+
+
+class SectionedChoiceState(ChoiceState):
+    def __init__(self, *args, sections: dict, separator: Optional[str] = "", **kwargs):
```
rudigiesler, line 260: Having `sections` be a `dict` could be a bit troublesome, since a `dict` doesn't have any ordering, and we probably want a fixed order for the sections. Maybe a `Sequence[Tuple[str, Sequence[Choice]]]` is better, eg. `[("section1", [Choice(...), ...]), ("section2", [Choice(...), ...])]`

vaccine/states.py
```diff
@@ -254,3 +254,32 @@ async def display(self, message: Message):
...
+
+        i = 1
+        for section, choices in self.sections.items():
+            lines.append(f"*{section}*")
+
+            for choice in choices:
+                lines.append(f"{i}. {choice.label}")
+                i += 1
+
+            lines.append(self.separator or "")
```
rudigiesler, line 283: Since `separator` is `Optional`, maybe we should take advantage of that, and leave the separator out if it's `None`. The way it is now, you can't not have a newline between sections.
ie.
```
if self.separator is not None:
    lines.append(self.separator)
```

rudigiesler (commented):
Overall looks good, just a few comments to make it a bit more flexible.

vaccine/states.py
```diff
@@ -254,3 +265,39 @@ async def display(self, message: Message):
...
+class SectionedChoiceState(ChoiceState):
+    def __init__(
+        self,
+        *args,
+        sections: Sequence[Tuple[str, Sequence[Choice]]],
+        separator: Optional[str] = None,
+        **kwargs,
+    ):
+        choices: List[Choice] = []
+        for section_choices in sections:
```
rudigiesler, line 279: Not a big one, but could probably do `for _, choices in sections` instead of the `section_choices[1]` below

vaccine/states.py
```diff
@@ -254,3 +265,39 @@ async def display(self, message: Message):
...
+        self.sections = sections
+        self.separator = separator
+
+    @property
+    def _display_choices(self) -> str:
+        lines = []
+
+        i = 1
+        for section in self.sections:
+            lines.append(f"{section[0]}")
```
rudigiesler, line 294: Could also do the `for section_name, choices in self.sections` , and then do a `lines.append(section_name)` here.

rudigiesler (commented):
:+1: Just a few very minor things, happy with this

PR #282, opened 2022-07-25
Get and send suggested content to Rapidpro

erikh360:
We gather the suggested content and save it on the Rapidpro Contact to send after 15 minutes of no response.

The choices are added to the state but not displayed so they'll still work after the user gets the suggested content options from rapidpro


Buhle79:
+1

PR #283, opened 2022-07-25
Cache random suggested content

erikh360:
The state gets executed again when user sends their response, this sops us from getting a different set of random articles


Buhle79:
+1

PR #309, opened 2022-08-17
Add suggested to page view data

erikh360:


yal/contentrepo.py
```diff
@@ -101,6 +101,9 @@ async def get_page_details(user, page_id, message_id):
                     "data__session_id": user.session_id,
```
fritzbrand, line 101: @erikh360 this double underscore naming convention in these variables is new to me.  I assume its just a way to "group" them, or am I missing something else here?

erikh360: It's how we tell the content repo to save it as extra details on the page view: https://github.com/praekeltfoundation/contentrepo/blob/main/home/models.py#L369-L372

We did it like this because the contentrepo doesn't have a concept of users etc

fritzbrand: 👍

PR #353, opened 2022-09-20
Fix for long list items

erikh360:


vaccine/states.py
```diff
@@ -269,7 +269,12 @@ async def display(self, message: Message):
             helper_metadata={
                 "button": self.button,
                 "sections": [
-                    {"rows": [{"id": c.label, "title": c.label} for c in self.choices]}
+                    {
+                        "rows": [
+                            {"id": c.label[:20], "title": c.label[:20]}
```
rudigiesler, line 274: So this might cause issues:
```
>>> a = "👍🏿"
>>> len(a)
2
>>> a[:1]
'👍'
```

But we do take this naive approach in the transport as well, so I guess it's fine for now. But we do need to have a better way of truncating for whatsapp, since emojis only count for a single character I believe

rudigiesler (approved):

vaccine/states.py
erikh360: hmm so my message is failing because of another reason

rudigiesler: No it could still be this. I see in the transport we limit the amount of items on the list, but not the text length for each of the items (just the length of the button text), so this could fix the issue

PR #363, opened 2022-09-27
Fix DeprecationWarnings by not setting variables on Sanic instances

jerith:
The first commit configures pytest to treat DeprecationWarnings as errors.
Following commits will fix the deprecated behaviour.

### DIFF
--- a/yal/tests/test_quiz.py
+++ b/yal/tests/test_quiz.py
@@ -17,11 +17,9 @@ def tester():
 async def contentrepo_api_mock(sanic_client):
     Sanic.test_mode = True
     app = Sanic("contentrepo_api_mock")
-    app.requests = []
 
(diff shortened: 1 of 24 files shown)


yal/tests/test_quiz.py
```diff
@@ -17,11 +17,9 @@ def tester():
 async def contentrepo_api_mock(sanic_client):
     Sanic.test_mode = True
     app = Sanic("contentrepo_api_mock")
-    app.requests = []
```
erikh360, line 20: Why isn't this one replaced with tstate?

jerith: Because nothing was using it and the tests passed without it. We can put it back if it's needed for something else, though.

erikh360: I see

erikh360 (approved):

PR #364, opened 2022-09-28
Add button after quiz question feedback for user to proceed

KaitCrawford:


yal/quiz.py
```diff
@@ -138,18 +138,27 @@ async def state_quiz_answer(self):
                 score += int(tag.replace("score_", ""))
         self.save_metadata("quiz_score", score)
 
+        self.save_metadata("quiz_sequence", metadata["quiz_sequence"] + 1)
+        return await self.go_to_state("state_quiz_answer_feedback")
+
+    async def state_quiz_answer_feedback(self):
+        metadata = self.user.metadata
+        page_id = metadata["selected_answer_id"]
+        error, page_details = await contentrepo.get_page_details(self.user, page_id, 1)
```
erikh360, line 147: Looks like we're doing this call in `state_quiz_process_answer` too, maybe we can put the body and image path in metadata then we don't have to call it again here

erikh360 (commented):
Looks good, just the one comment.

erikh360 (approved):
🍰

PR #365, opened 2022-09-28
Fix aaq double content send

erikh360:


yal/askaquestion.py
```diff
@@ -243,8 +243,10 @@ async def state_is_question_answered(self):
         inbound_id = self.user.metadata["inbound_id"]
         feedback_secret_key = self.user.metadata["feedback_secret_key"]
         feedback_type = "positive" if feedback_answer == "yes" else "negative"
+        # TODO: figure out where this id should come from
```
erikh360, line 246: This will come from the API, same as `feedback_secret_key` and `inbound_id`

PR #368, opened 2022-09-29
Add page feedback calls

erikh360:


Buhle79:
image

Buhle79 (approved):

yal/askaquestion.py
```diff
@@ -163,7 +163,16 @@ async def state_display_results(self):
     async def state_set_aaq_timeout_2(self):
         chosen_answer = self.user.answers.get("state_display_results")
 
+        inbound_id = self.user.metadata["inbound_id"]
+        feedback_secret_key = self.user.metadata["feedback_secret_key"]
+
         if chosen_answer == "more":
+            error = await aaq_core.add_feedback(
+                feedback_secret_key, inbound_id, "negative", page="1"
+            )
```
KaitCrawford, line 172: @erikh360 For my own understanding: what are we doing here? If the user chooses "more" then we're returning "negative" feedback for the page?

erikh360: Yes check some of the comments here: https://miro.com/app/board/uXjVO9B3_Ek=/?moveToWidget=3458764534703136201&cot=14

So moving from page 1 -> 2 means page 1 isn't good enough, same if they request a call on page 2

Themba has all the context 🧠

KaitCrawford: Right, that makes sense 🙇 thanks

PR #374, opened 2022-10-05
YAL: update content for aaq flows

KaitCrawford:


yal/askaquestion.py
```diff
@@ -52,12 +52,11 @@ async def state_aaq_start(self):
...
             "\n".join(
                 [
-                    "🙋🏿‍♂️ *QUESTIONS?*",
-                    "Ask A Question",
+                    "🙋🏿‍♂️ QUESTIONS? / *Ask A Question*",
                     "-----",
                     "",
-                    "🙍🏾‍♀️*That's what I'm here for!*",
-                    "*Just type your Q and hit send* .🙂.",
+                    "[persona_emoji] *That's what I'm here for!*",
```
rudigiesler, line 58: Should this be a to do? I don't see this being replaced with the emoji anywhere

rudigiesler (commented):
Just one question, otherwise looks good!

yal/askaquestion.py
KaitCrawford: Nope. this is the placeholder we're using. my next ticket is to replace the placeholder with an emoji across all messages.

PR #375, opened 2022-10-06
Feature/yal 254 bring flows in line with designs

KaitCrawford:
Additional formatting and changes needed to the YAL AAQ states

### DIFF
--- a/yal/askaquestion.py
+++ b/yal/askaquestion.py
@@ -141,10 +141,10 @@ class Application(BaseApplication):
             ]
             # Add next page option if there is one
             if self.user.metadata.get("next_page_url"):
-                choices.append(Choice("more", "See more options"))
                 question_list.extend(
-                    ["or", f"*{len(answers)+1}*. See more options", ""]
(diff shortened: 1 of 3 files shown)


yal/askaquestion.py
```diff
@@ -165,14 +163,20 @@ async def state_display_results(self):
                     "answer, reply with the number of the topic you're interested "
                     "in.",
                     "",
-                    get_display_choices(choices),
+                    get_display_choices(choices, bold_numbers=True),
+                    "",
+                    "or",
+                    f"*{len(answers)+1}*. Back to first list",
+                    f"*{len(answers)+2}*. Talk to a counsellor",
```
rudigiesler, line 170: Shouldn't this rather be `len(choices)`? I think it might be the same at this point, but I think it would make more sense for this to match what is being passed to `get_display_choices`

rudigiesler (commented):
Just one small comment, rest looks good :)

yal/askaquestion.py
KaitCrawford: Done.

PR #376, opened 2022-10-10
Handle persona variables

KaitCrawford:
In order to give the user the feel of a custom persona we need to replace variables in the copy with their chosen values.

This adds two methods to the main application to catch the different message types and send them to a util method to do the variable replacement.

### DIFF
--- a/yal/utils.py
+++ b/yal/utils.py
@@ -70,7 +70,7 @@ def normalise_phonenumber(phonenumber):
 def replace_persona_fields(text, metadata={}):
     for key in PERSONA_FIELDS:
         if key in metadata:
-            return text.replace(f"[{key}]", metadata[key])
+            text = text.replace(f"[{key}]", metadata[key])
         else:
             text = text.replace(f"[{key}]", PERSONA_DEFAULTS[key])
     return text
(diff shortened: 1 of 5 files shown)


yal/utils.py
```diff
@@ -64,3 +66,12 @@ def normalise_phonenumber(phonenumber):
         return phonenumbers.format_number(pn, phonenumbers.PhoneNumberFormat.E164)
     except (phonenumbers.phonenumberutil.NumberParseException, AssertionError):
         raise ValueError("Invalid phone number")
+
+
+def replace_persona_fields(text, metadata={}):
+    for key in PERSONA_FIELDS:
+        if key in metadata:
+            text = text.replace(f"[{key}]", metadata[key])
```
rudigiesler, line 74: I'm guessing it's too late to rather use something like https://docs.python.org/3.8/library/string.html#string.Template? Since it uses a different placeholder format.

rudigiesler (approved):

yal/utils.py
KaitCrawford: yeah, unfortunately it is a bit too late. Codie, Erik and I discussed it and chose this format since it's applicable across platforms (python framework vs rapidpro) and doesn't clash with the Turn template. But we didn't consider this option. We also didn't take into account the clash with Markdown's formatting.
but unfortunately the content has already been formatted this way. I think we can reopen this format conversation after the mvp

PR #378, opened 2022-10-12
YAL ContentRepo Banner Messages

rudigiesler:
If there are any messages tagged with "banner" in the content repo, we want to send that content as additional messages after the main menu

### DIFF
--- a/vaccine/tests/test_worker.py
+++ b/vaccine/tests/test_worker.py
@@ -407,7 +407,7 @@ async def test_answer_worker_push_results_user_error(
     )
 
     # wait for worker to log error
-    for _ in range(10):
+    for _ in range(50):
         if len(flow_results_mock_server.tstate.requests) == 3:
             break
         await sleep(0.1)
--- a/yal/contentrepo.py
+++ b/yal/contentrepo.py
@@ -1,8 +1,10 @@
 import logging
+from typing import Any, Dict, List, Tuple
 from urllib.parse import urljoin
 
 import aiohttp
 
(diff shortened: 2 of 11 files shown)


vaccine/tests/test_worker.py
```diff
@@ -407,7 +407,7 @@ async def test_answer_worker_push_results_user_error(
     )
 
     # wait for worker to log error
-    for _ in range(10):
+    for _ in range(50):
```
rudigiesler, line 410: Make this test more reliable

yal/contentrepo.py
```diff
@@ -171,7 +181,7 @@ async def get_page_details(user, page_id, message_id, suggested=False):
             except HTTP_EXCEPTIONS as e:
                 if i == 2:
                     logger.exception(e)
-                    return True, []
+                    return True, {}
```
rudigiesler, line 184: There wasn't any typing on these functions, so I added some, which revealed some issues that I also fixed

yal/tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+from unittest import mock
+
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="package")
+def mock_sleep():
+    with mock.patch("asyncio.sleep"):
+        yield
```
rudigiesler, line 9: This is to speed up tests, so that we don't need to wait for all the `sleep` calls

KaitCrawford (approved):
👍 Looks great!

PR #380, opened 2022-10-12
Fix the use of message_id for contentrepo

rudigiesler:
The message_id param is not the ID of the inbound message for tracking purposes. It is which message out of all the messages in the piece of content we want details for. So we need to go through all the messages for each content page, to get all the messages to send for the banner.


KaitCrawford (approved):
👍

PR #385, opened 2022-10-13
content feedback

rudigiesler:
Adds the routing, as well as the flow, for handling responses when we send a push message asking the user to give their feedback on the content that they just viewed

### DIFF
--- a/yal/content_feedback_survey.py
+++ b/yal/content_feedback_survey.py
@@ -2,6 +2,7 @@ from vaccine.base_application import BaseApplication
 from vaccine.states import Choice, EndState, FreeText, WhatsAppButtonState
 from vaccine.utils import get_display_choices
 from yal import rapidpro, utils
+from yal.askaquestion import Application as AAQApplication
 
 
 class ContentFeedbackSurveyApplication(BaseApplication):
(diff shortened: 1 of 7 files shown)


yal/content_feedback_survey.py
```diff
@@ -0,0 +1,156 @@
...
+    START_STATE = "state_content_feedback_survey_start"
+
+    async def state_content_feedback_survey_start(self):
+        msisdn = utils.normalise_phonenumber(self.inbound.from_addr)
+        whatsapp_id = msisdn.lstrip("+")
+        # Reset this, so that we only get the survey once after a push
+        await rapidpro.update_profile(whatsapp_id, {"feedback_survey_sent": ""})
+
+        keyword = utils.clean_inbound(self.inbound.content)
+        print(keyword)
```
KaitCrawford, line 18: stray Print

KaitCrawford (approved):
👍

KaitCrawford:
@rudigiesler it looks like there have been some minor copy changes to the Miro board for these states

yal/mainmenu.py
```diff
@@ -217,6 +218,18 @@ async def next_(choice: Choice):
         )
 
     async def state_contentrepo_page(self):
+        msisdn = utils.normalise_phonenumber(self.inbound.from_addr)
+        whatsapp_id = msisdn.lstrip("+")
+        timestamp = datetime.now() + timedelta(minutes=10)
```
KaitCrawford, line 223: There is a `get_current_datetime` method in yal.utils
maybe we should use that?

KaitCrawford (approved):
👍

PR #386, opened 2022-10-16
Offer AAQ at onboarding completion

KaitCrawford:
In order to redirect to AAQ with different message text I created a method to forward given kwargs along when trying to "go_to_state"

### DIFF
--- a/vaccine/base_application.py
+++ b/vaccine/base_application.py
@@ -39,28 +39,18 @@ class BaseApplication:
         )
         self._ = self.translation.gettext
 
-    async def get_current_state(self):
-        if not self.state_name:
-            self.state_name = self.START_STATE
-        state_func = getattr(self, self.state_name)
-        return await state_func()
-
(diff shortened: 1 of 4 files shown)


vaccine/base_application.py
```diff
@@ -52,6 +52,16 @@ async def go_to_state(self, name):
...
+    async def go_to_state_with_kwargs(self, name, **kw):
+        """
+        Go to another state and have it process the user message instead
+        """
+        self.state_name = name
+        if not self.state_name:
+            self.state_name = self.START_STATE
+        state_func = getattr(self, self.state_name)
+        return await state_func(**kw)
+
```
KaitCrawford, line 64: This is a slight variation on the method above which sends the user to a state

KaitCrawford: 👍 Updated

bearnard:
when green 👍

PR #388, opened 2022-10-17
Add custom persona to various features

KaitCrawford:


bearnard:
Without context, LGTM 👍

yal/pleasecallme.py
```diff
@@ -182,24 +183,22 @@ async def state_open_hours(self):
         )
 
     async def state_in_hours_greeting(self):
-        await self.worker.publish_message(
-            self.inbound.reply(
-                self._("👩🏾 *Say no more—I'm on it!*\n☝🏾 Hold tight just a sec..."),
-            )
+        await self.publish_message(
```
bearnard, line 186: what's the difference between worker publish_message and just self. publish_message ?

KaitCrawford: self.publish_message calls the thing that replaces the placeholders with the values and then calls worker.publish_message
https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/main.py#L145

PR #390, opened 2022-10-19
fix content feedback trigger

rudigiesler:
This won't only occur on new sessions, this might occur in exiting sessions, so we need to move it to the process_message function

### DIFF
--- a/yal/onboarding.py
+++ b/yal/onboarding.py
@@ -97,17 +97,15 @@ class Application(BaseApplication):
         return await self.go_to_state("state_profile_intro")
 
     async def state_profile_intro(self):
-        msg = (
-            self._(
-                "\n".join(
-                    [
-                        "*[persona_emoji] Great! I'm just going to ask you a few "
-                        "quick questions now.*",
-                    ]
-                )
-            ),
+        msg = self._(
+            "\n".join(
+                [
+                    "*[persona_emoji] Great! I'm just going to ask you a few "
+                    "quick questions now.*",
+                ]
+            )
         )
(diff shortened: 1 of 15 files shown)


KaitCrawford (commented):
@rudigiesler are all the changes to tests necessary because "1" and "2" are in the feedback keywords? Is there any way we can avoid that?

yal/onboarding.py
```diff
@@ -97,17 +97,15 @@ async def state_save_persona_emoji(self):
...
+        msg = self._(
+            "\n".join(
+                [
+                    "*[persona_emoji] Great! I'm just going to ask you a few "
+                    "quick questions now.*",
+                ]
+            )
         )
-        self.publish_message(msg)
+        await self.publish_message(msg)
```
rudigiesler, line 108: Unrelated, but it was showing a warning, so I fixed it

rudigiesler:
@KaitCrawford Yes, `1`, `2`, and `yes` are fairly common responses. We could possible remove `yes`, because that isn't explicitly mentioned, but `1` and `2` are in the content of the message sent to the user, so the only way we can avoid that is to change the content.

KaitCrawford:
> @KaitCrawford Yes, `1`, `2`, and `yes` are fairly common responses. We could possible remove `yes`, because that isn't explicitly mentioned, but `1` and `2` are in the content of the message sent to the user, so the only way we can avoid that is to change the content.

bah humbug! ☹️

KaitCrawford (approved):
👍

PR #392, opened 2022-10-20
Add flows for the whatsapp facebook crossover feedback flows

megh-k:
No implementation just yet. just the flows with content and tests

### DIFF
--- a/yal/tests/test_wa_fb_crossover_feedback.py
+++ b/yal/tests/test_wa_fb_crossover_feedback.py
@@ -5,7 +5,6 @@ import pytest
 from sanic import Sanic, response
 
 from vaccine.testing import AppTester, TState, run_sanic
-from vaccine.models import Message
 from yal import config
 from yal.wa_fb_crossover_feedback import Application
 
--- a/yal/tests/test_wa_fb_crossover_feedback.py
+++ b/yal/tests/test_wa_fb_crossover_feedback.py
@@ -111,7 +111,7 @@ async def test_state_fb_hot_topic_helpful(tester: AppTester, rapidpro_mock):
     tester.setup_state("state_saw_recent_facebook")
     await tester.user_input("It was helpful")
     assert tester.application.answer_events[0].response == "helpful"
-    tester.assert_state(None)
+    tester.assert_state("state_fb_hot_topic_helpful")
 
 
 @pytest.mark.asyncio
@@ -119,7 +119,7 @@ async def test_state_fb_hot_topic_helpful_learnt_new(tester: AppTester, rapidpro
     tester.setup_state("state_saw_recent_facebook")
     await tester.user_input("Learnt something new")
     assert tester.application.answer_events[0].response == "learnt new"
-    tester.assert_state(None)
+    tester.assert_state("state_fb_hot_topic_helpful")
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


yal/tests/test_wa_fb_crossover_feedback.py
```diff
@@ -0,0 +1,155 @@
...
+    tester.setup_state("state_wa_fb_crossover_feedback")
+    await tester.user_input("No, I didn't")
+    tester.assert_state(None)
+    assert len(rapidpro_mock.tstate.requests) == 1
+    assert [r.path for r in rapidpro_mock.tstate.requests] == ["/api/v2/contacts.json"]
+
+    post_request = rapidpro_mock.tstate.requests[0]
+
+    engaged_on_facebook = json.loads(post_request.body.decode("utf-8"))["fields"][
+        "engaged_on_facebook"
```
KaitCrawford, line 69: pls check both fields that you're setting in this test.
You can use this test for an example of how to do that, complete with a mocked datetime
https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/tests/test_pleasecallme.py#L495

yal/tests/test_wa_fb_crossover_feedback.py
```diff
@@ -0,0 +1,155 @@
...
+
+    assert engaged_on_facebook is True
+
+
+@pytest.mark.asyncio
+async def test_state_fb_hot_topic_helpful(tester: AppTester, rapidpro_mock):
+    tester.setup_state("state_saw_recent_facebook")
+    await tester.user_input("It was helpful")
+    assert tester.application.answer_events[0].response == "helpful"
+    tester.assert_state(None)
```
KaitCrawford, line 114: By my calculation the user should be in `state_fb_hot_topic_helpful` 
If they're ending in "None" then there's a problem.

yal/wa_fb_crossover_feedback.py
```diff
@@ -0,0 +1,257 @@
...
+from yal.askaquestion import Application as AskaQuestionApplication
+from yal.pleasecallme import Application as PleaseCallMeApplication
+from yal.utils import get_generic_error
+
+logger = logging.getLogger(__name__)
+
+
+class Application(BaseApplication):
+    START_STATE = "state_wa_fb_crossover_feedback"
+
```
KaitCrawford, line 16: Please add another state here at the start that is a carbon copy of this one (but the name mustn't clash and it must take the user to your start state)
https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/content_feedback_survey.py#L11
It's needed to let RapidPro know that the user has been offered this survey so it doesn't get sent again

yal/tests/test_wa_fb_crossover_feedback.py
```diff
@@ -0,0 +1,155 @@
...
+
+    engaged_on_facebook = json.loads(post_request.body.decode("utf-8"))["fields"][
+        "engaged_on_facebook"
+    ]
+
+    assert engaged_on_facebook is True
+
+
+@pytest.mark.asyncio
+async def test_state_fb_hot_topic_helpful(tester: AppTester, rapidpro_mock):
```
KaitCrawford, line 110: Do you still need the `rapidpro_mock` in the signature?

KaitCrawford: Nevermind! 😅

yal/tests/test_wa_fb_crossover_feedback.py
megh-k: @KaitCrawford I thought it might be going into None because something is wrong with how I have done the talk to a counsellor and aaq but I dont know, I've been over it so many times and I cant figure it out

yal/tests/test_wa_fb_crossover_feedback.py
megh-k: The miro board flow doesnt update the recommended content if the user did not see the facebook post, so I didnt update that field in that flow, should I?

KaitCrawford: No that's fine

PR #394, opened 2022-10-20
Fix # symbol keyword

rudigiesler:
Before, it was being stripped out before being compared with the list of keywords, so it would never match. This ensures that it doesn't get stripped out


KaitCrawford (approved):
👍

PR #395, opened 2022-10-20
servicefinder result changes

rudigiesler:
- Add google maps URL to servicefinder results
- Don't show phone number if there isn't one


KaitCrawford (approved):
👍 nice!

PR #397, opened 2022-10-20
Add articles with help_submenu tag to main menu

KaitCrawford:
There is some content from the content repo that should be displayed under the "Need Help?" section of the main menu. This pulls it

### DIFF
--- a/yal/mainmenu.py
+++ b/yal/mainmenu.py
@@ -72,6 +72,10 @@ class Application(BaseApplication):
     async def state_mainmenu(self):
         self.save_metadata("current_menu_level", 0)
 
+        error, submenu_choices = await contentrepo.get_choices_by_tag("help_submenu")
+        if error:
+            return await self.go_to_state("state_error")
+
         sections = [
             (
                 "*🏥 NEED HELP?*",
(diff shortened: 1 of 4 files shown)


yal/mainmenu.py
```diff
@@ -72,6 +72,10 @@ async def state_pre_mainmenu(self):
     async def state_mainmenu(self):
         self.save_metadata("current_menu_level", 0)
 
+        error, submenu_choices = await contentrepo.get_choices_by_tag("help_submenu")
+        if error:
+            return await self.go_to_state("state_error")
```
rudigiesler, line 77: This affects the whole service, so not something now, but maybe for when adding things in future. We should be `logger.error(...)` when this happens, so that we can get a sentry notification when something is broken. Right now there's nothing in the logs when errors occur, and that can make finding and fixing errors difficult.

rudigiesler (approved):

yal/mainmenu.py
KaitCrawford: Exceptions for this are already raised and logged in the contentrepo method though
https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/contentrepo.py#L69

rudigiesler: :+1: missed that

PR #400, opened 2022-10-21
AAQ Feedback flow updates, implement questions not answered, back and questions answered

megh-k:


yal/askaquestion.py
```diff
@@ -254,26 +255,31 @@ async def state_display_content(self):
...
-                Choice("yes", "Yes"),
-                Choice("no", "No"),
-            ],
+            choices=choices,
             error=self._(get_generic_error()),
-            next="state_is_question_answered",
+            next={
+                "yes": "state_is_question_answered",
+                "back to list": "state_display_results",
+                "no": "state_no_question_not_answered",
```
KaitCrawford, line 281: the "no" response must also please go to `state_is_question_answered` first.
This is so that we can send the positive or negative feedback back to the model
So the bottom of `state_is_question_answered` should direct to `state_no_question_not_answered` instead

yal/tests/test_askaquestion.py
```diff
@@ -413,51 +427,76 @@ async def test_state_display_content_question_not_answered(
...
+        ]
+    )
 
-    assert len(aaq_mock.tstate.requests) == 1
-    request = aaq_mock.tstate.requests[0]
-    assert json.loads(request.body.decode("utf-8")) == {
-        "feedback": {"feedback_type": "negative", "faq_id": "1"},
-        "feedback_secret_key": "feedback-secret-key",
-        "inbound_id": "inbound-id",
-    }
```
KaitCrawford, line 432: These lines need to still be here because we still need to do this for negative responses (this makes sure we're updating the model with the negative response feedback)

yal/askaquestion.py
KaitCrawford: `back to list` should also go to state_is_question_answered` for the same reason. you can do this split based on the users input at the end of that

yal/tests/test_askaquestion.py
```diff
@@ -505,13 +544,26 @@ async def test_state_handle_timeout_handles_type_2_yes(
...
+            "1. No changes",
+            "2. Yes, I have a change",
+            "",
+            "-----",
+            "*Or reply:*",
+            BACK_TO_MAIN,
+            GET_HELP,
+        ]
     )
+    tester.assert_message(message)
```
KaitCrawford, line 566: After the changes requested this test and the one below will probably fail because they need the `aaq_mock` parameter and you should test that we're sending the correct feedback on the model

yal/tests/test_askaquestion.py
```diff
@@ -526,7 +578,108 @@ async def test_state_handle_timeout_handles_type_2_no(tester: AppTester, rapidpr
...
+            "*Or reply:*",
+            BACK_TO_MAIN,
+            GET_HELP,
+        ]
+    )
+    tester.assert_message(message)
+
+
+@pytest.mark.asyncio
+async def test_state_save_aaq_priority_feedback(tester: AppTester, rapidpro_mock):
```
KaitCrawford, line 659: this is an odd name for this test? 🤔

yal/tests/test_main.py
```diff
@@ -256,8 +257,24 @@ async def test_callback_check_response_to_handler(tester: AppTester):
...
+            "for you next time.",
+            "",
+            "_Just type and send your question again, now._" "",
+            "-----",
+            "*Or reply:*",
+            BACK_TO_MAIN,
+            GET_HELP,
+        ]
+    )
+    tester.assert_message(message)
```
KaitCrawford, line 277: After the changes you'll also need to add `aaq_mock` to this test as well (and the definition for it might not exist in this file yet 😬 ) but you can leave out the tests to check the request that went through

KaitCrawford (commented):
One awkward change needed and I've tried to comment where I think it'll cause trouble for the tests

yal/tests/test_askaquestion.py
megh-k: Oh this was after I misunderstood a part of the miro board, forgot to take it out, thanks for catching

KaitCrawford (approved):
👍 🍰 🍷

PR #403, opened 2022-10-21
Service finder feedback flow

rudigiesler:
- Add servicefinder feedback flow
- Trigger survey from servicefinder flow
- Add service finder feedback flow triggers to main

### DIFF
--- a/yal/servicefinder_feedback_survey.py
+++ b/yal/servicefinder_feedback_survey.py
@@ -1,0 +1,758 @@
+from datetime import datetime, timedelta
+
+from vaccine.base_application import BaseApplication
+from vaccine.states import Choice, FreeText, WhatsAppButtonState
+from vaccine.utils import get_display_choices
+from yal import rapidpro, utils
+from yal.askaquestion import Application as AAQApplication
+from yal.pleasecallme import Application as PCMApplication
+from yal.servicefinder import Application as SFApplication
+
+
+class ServiceFinderFeedbackSurveyApplication(BaseApplication):
+    START_STATE = "state_servicefinder_feedback_survey_start"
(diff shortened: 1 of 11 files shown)


KaitCrawford (approved):
👍 regardless of what you decide about my comment

yal/servicefinder_feedback_survey.py
```diff
@@ -179,10 +179,15 @@ async def state_servicefinder_feedback_confirmation(self):
...
         whatsapp_id = utils.normalise_phonenumber(self.inbound.from_addr).lstrip("+")
+        # Reset this, so that we only get the survey once after a push
+        await rapidpro.update_profile(whatsapp_id, {"feedback_survey_sent_2": ""})
         error, fields = await rapidpro.get_profile(whatsapp_id)
         if error:
             return await self.go_to_state("state_error")
-        feedback_datetime = datetime.fromisoformat(fields["feedback_timestamp"])
+        feedback_datetime = (
+            datetime.fromisoformat(fields["feedback_timestamp_2"])
+            - self.CALLBACK_2_DELAY
```
KaitCrawford, line 189: I don't understand why we're subtracting this here now?

rudigiesler: This is an issue I picked up when doing this change. The `feedback_timestamp_2` is the timestamp of when the message goes out. So it will be today's date, which is not when they searched for a clinic. They searched for a clinic 2 weeks ago, and we're following up with them now.

KaitCrawford: Ah I see. thanks for clarifying.

KaitCrawford (approved):
👍

PR #415, opened 2022-10-25
change feedback survey response logic

rudigiesler:
- Only check rapidpro contact if feedback_timestamp is within a minute of current time
- Ensure that the feedback state gets called for both new and existing sessions

### DIFF
--- a/yal/askaquestion.py
+++ b/yal/askaquestion.py
@@ -44,7 +44,14 @@ class Application(BaseApplication):
     START_STATE = "state_aaq_start"
     TIMEOUT_RESPONSE_STATE = "state_handle_timeout_response"
     TRIGGER_KEYWORDS = {
-        "1", "yes ask again", "yes", "2", "no i m good", "nope", "3", "no go back to list"
+        "1",
+        "yes ask again",
+        "yes",
+        "2",
+        "no i m good",
+        "nope",
+        "3",
+        "no go back to list",
     }
 
     async def state_aaq_start(self, question=None, buttons=None):
@@ -528,9 +535,7 @@ class Application(BaseApplication):
             # Get it to display the message, instead of having this state try to
             # match it to a keyword
             self.inbound.session_event = Message.SESSION_EVENT.NEW
-            return await self.go_to_state(
-                "state_aaq_timeout_unrecognised_option"
-            )
+            return await self.go_to_state("state_aaq_timeout_unrecognised_option")
 
     async def state_aaq_timeout_unrecognised_option(self):
         choices = [
--- a/yal/askaquestion.py
+++ b/yal/askaquestion.py
@@ -43,7 +43,7 @@ def get_aaq_api():
 class Application(BaseApplication):
     START_STATE = "state_aaq_start"
     TIMEOUT_RESPONSE_STATE = "state_handle_timeout_response"
-    TRIGGER_KEYWORDS = {
+    AAQ_FEEDBACK_TRIGGER_KEYWORDS = {
         "1",
         "yes ask again",
         "yes",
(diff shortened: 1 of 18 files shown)


yal/main.py
```diff
@@ -102,33 +90,49 @@ async def process_message(self, message):
...
+
+    async def get_feedback_state(self):
+        """
+        If the user needs to be in a feedback state, send return that state name,
+        otherwise return None
+        """
+        feedback_timestamp = self.user.metadata.get("feedback_timestamp")
+        feedback_in_time = feedback_timestamp and datetime.fromisoformat(
+            feedback_timestamp
+        ) > get_current_datetime() - timedelta(minutes=1)
```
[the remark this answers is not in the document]
KaitCrawford: oh, nevermind. My brain got there now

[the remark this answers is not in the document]
rudigiesler: Yeah it is, I split it out instead of having a giant if statement, although maybe I should split it out further.

yal/askaquestion.py
```diff
@@ -502,16 +516,57 @@ async def state_handle_timeout_response(self):
...
-            "feedback_survey_sent": "",
-        }
+        self.save_metadata("feedback_timestamp", "")
+        data = {"feedback_survey_sent": "", "feedback_timestamp": ""}
         error = await rapidpro.update_profile(whatsapp_id, data)
         if error:
             return await self.go_to_state("state_error")
 
         timeout_type_sent = fields.get("feedback_type")
+        self.save_metadata("feedback_type", timeout_type_sent)
```
KaitCrawford, line 526: Do we ever use `feedback_type` from the metadata?

yal/askaquestion.py
```diff
@@ -502,16 +516,57 @@ async def state_handle_timeout_response(self):
...
+        return WhatsAppButtonState(
+            self,
+            question=question,
+            choices=choices,
+            error=get_generic_error(),
+            next={
+                "feedback": feedback_state,
+                # hardcode to prevent circular import
+                "mainmenu": "state_pre_mainmenu",
+                "aaq": self.START_STATE,
```
KaitCrawford, line 570: We've been finding that this doesn't work in production. The user ends up getting sent to `state_start` instead of the state we want

KaitCrawford (commented):
Just a small comment

yal/askaquestion.py
rudigiesler: This is just specifically for the AAQ flow.

I didn't want to make too many changes to how the flows work, and this flow works slightly differently to others. Instead of doing the routing to the correct state in the main flow, AAQ routes all types to one state, and that state sends the user to the correct state depending on the feedback type.

I didn't want to change that much, but I did need to now be able to store that and use it later, so that the error state could also route to the correct state (or I guess I could've created 2 error states). So this is only used so that when you say "go back to survey" in the error state, it knows which state to send you to.

yal/askaquestion.py
rudigiesler: Ah of course, because we're using inheritance to split this flow into multiple flows (also a bit of a hack). I'll hardcode it

yal/askaquestion.py
KaitCrawford: ah ok! no worries then 👍

yal/askaquestion.py
```diff
@@ -92,7 +92,7 @@ async def state_coming_soon(self):
         return EndState(
             self,
             self._("Coming soon..."),
-            next=self.START_STATE,
+            next="state_aaq_start",
```
KaitCrawford, line 95: This is a nit-picking comment since this state shouldn't be reachable: won't this send them into an infinite loop? because they get here from `state_aaq_start` if the url isn't configured

KaitCrawford (approved):
Small comment about something that isn't important

yal/askaquestion.py
rudigiesler: Ah true, I'll change this one back and merge, since it's a small change

PR #421, opened 2022-10-25
Fix service finder location province sets to skip

megh-k:


yal/servicefinder.py
```diff
@@ -559,6 +559,12 @@ async def state_save_location(self):
         return await self.go_to_state("state_category_lookup")
 
     async def state_province(self):
+        def _next(choice: Choice):
+            if choice.value == "skip":
+                return "state_validate_full_address_error"
```
KaitCrawford, line 564: We want to allow them to skip this field. So they still need to go on to the next state and then where we store the data we should ensure we don't store "skip"

megh-k: Ah ok, so street address and no province is fine?

KaitCrawford: yup

KaitCrawford (approved):
👍

PR #430, opened 2022-10-26
Add reset feedback timestamp keyword

rudigiesler:
This is a keyword in order to speed up QA. It resets the feedback timestamp to the current time, so that the feedback flows can be tested immediately, instead of waiting for the delay time


KaitCrawford (approved):
👍

PR #439, opened 2022-11-07
Update pytest and pytest-cov

jerith:
This was supposed to remove `py` as a dependency and avoid a security alert that isn't actually a problem, but `locust` still depends on it through `pyzmq`.


rudigiesler (approved):
`locust` is used for performance testing, but not in any automated/CI way. It's just used in https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/vaccine/tests/vaccine_reg_loadtest.py , so I'm not too worried about it, if I understand the description of the security issue

jerith:
I'm not concerned about the issue (we don't use `svn`, which means we'll never hit the problematic code), but updating `pytest` is an easy way to make it go away... unless something else also depends on it. We can tell dependabot to ignore it in this repo.

PR #447, opened 2022-11-14
Update survey flow triggers

erikh360:


yal/onboarding.py
```diff
@@ -109,13 +109,6 @@ async def state_profile_intro(self):
         return await self.go_to_state("state_age")
 
     async def state_age(self):
-        msisdn = utils.normalise_phonenumber(self.inbound.from_addr)
-        if (
-            config.SEGMENTATION_SURVEY_ACTIVE
-            or msisdn in config.SEGMENTATION_SURVEY_ALLOWED
-        ):
```
rudigiesler, line 116: Do we maybe want to remove these configs if we're no longer using them?

rudigiesler (approved):

yal/onboarding.py
erikh360: Ah yes, I forgot about them

PR #449, opened 2022-11-14
Label airtime failure reports for helpdesk

erikh360:


KaitCrawford (commented):
Why are we using Turn for this? Why not RapidPro or something else? are we using Turn for anything else on this project (besides as the BSP)?

erikh360:
@KaitCrawford I was hoping this is something a helpdesk operator could pick up and action, but I see we don't have that as yet. Let me see what else we can do

KaitCrawford:
@erikh360 I agree that it's something a human should investigate but I'm concerned we're unnecessarily increasing the number of tools people have to use on a day-to-day basis to manage the project.
Although I'm not sure an airtime error is something that anyone other than an engineer would be able to investigate as things stand

erikh360:
@KaitCrawford we already get DTone emails when airtime transfers fail and engineers can action those. My thinking here was more around people taking chances for free airtime. Helpdesk operators can check the chat history to see if they actually did the survey. but now I'm thinking the only way this label is added is at the end of the survey so maybe useless. 

I'm going to think through this a bit more

PR #450, opened 2022-11-15
Add override answer name to states

erikh360:


yal/segmentation_survey.py
```diff
@@ -153,11 +155,9 @@ async def state_survey_process_answer(self):
 
         section = metadata.get("segment_section", 1)
         current_question = metadata.get("segment_question")
-        answer = answers.get("state_survey_question")
+        answer = answers.get(current_question)
         question_number = metadata.get("segment_question_nr", 1)
 
-        self.save_answer(current_question, answer)
```
erikh360, line 159: This caused each answer to be saved twice in the flow results app.

vaccine/states.py
```diff
@@ -73,6 +73,7 @@ def __init__(
         error_header: Optional[str] = None,
         buttons: Optional[List[Choice]] = None,
         helper_metadata: Optional[dict] = None,
+        override_answer_name: Optional[str] = None,
```
rudigiesler, line 76: I'm wondering if maybe we should rather have this be `save_answer`, defaulting to `True`? And setting to `False` means that the answer isn't saved. I feel like that would be more useful in the general case. I'm happy with either solution though

rudigiesler (approved):

vaccine/states.py
erikh360: Hmm I like that idea, not one of the options I thought of.

erikh360: @rudigiesler If we don't save the answer we can't retrieve it later to save with correct name: https://github.com/praekeltfoundation/vaccine-eligibility/pull/450/files#diff-09324e18712c4ef1ef3e57882382696fc7cc0aa5c52456ec3c392d4a845f8595L156

rudigiesler: Ah right, didn't think of that, we've tied flow results and answers together, so you can't have one but not the other. This solution seems best then

PR #452, opened 2022-11-15
Fix update breadcrumbs in both service finder and pleasecallme flows

megh-k:
Updated breadcrumbs to match the current state of the miroboard. 
Related ticket: https://praekeltorg.atlassian.net/browse/YAL-374

### DIFF
--- a/yal/servicefinder.py
+++ b/yal/servicefinder.py
@@ -69,8 +69,7 @@ class Application(BaseApplication):
         question = self._(
             "\n".join(
                 [
-                    "🏥 Find Clinics and Services",
-                    "*Get help near you*",
+                    "NEED HELP? / *Find clinics and services*",
                     "-----",
                     "",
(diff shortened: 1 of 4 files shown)


yal/servicefinder.py
```diff
@@ -251,8 +247,8 @@ async def next_(choice: Choice):
         question = self._(
             "\n".join(
                 [
-                    "🏥 Find Clinics and Services",
-                    metadata["servicefinder_breadcrumb"],
+                    "NEED HELP? / Find clinics and services / "
+                    f"{metadata['servicefinder_breadcrumb']}",
```
erikh360, line 251: Does this need to be an fstring if its only the one value?

megh-k: @erikh360  Yeah it needs to be on a separate line because the line is too long but in the message it has to be on the same line

PR #457, opened 2022-11-17
Add emergency keywords that take users to the please call me flow

megh-k:
SXD has provided us with a sheet with the emergency keywords that need to be supported. I have included the csv exported, so that it is easy for us to update based on the sheet SXD is using.

After chatting with Chuk, they would like us to accommodate for variations of these words/phrases. I spoke to Rudi about fuzzy search and then implemented a fuzzy search with a cutoff score of 75, this unfortunately does catch some other keywords, which makes me nervous, but the check is done first, so if those keywords are in other lists, they will be overwritten. Where this really makes me nervous is misspellings of the other keywords _might_ incorrectly categorize a keyword. 

I think there are three solutions to this, 
1. tighten the acceptable score and miss out on catching phrase variations and only catch misspellings 
2. request a significantly more substantial list of phrases from SXD and tighten the scoring 
3. implement fuzzy search for all the keywords (I think this is a lovely idea, but I dont know how many problems it may cause, but I'd be happy to investigate as it will account for any typos users send in)

I have also tabulated the matches below for a positive match on a score of 75 or above 

| keyword | match |  match ratio |
|---|---|---:|
| hi | hit | 80.0 |
| depressed | depressed | 100.0 |
| i want to die | i want to die | 100.0 |
| edpressed | depressed | 88.9 |
| i am depressed | depressed | 78.3 |
| help | help | 100.0 |
| hi | hit | 80.0 |
| heita | hit | 75.0 |

For a score of 85 and above, which catches typos and misses phrases unless they are in the csv:

| keyword | match |  match ratio |
|---|---|---:|
| depressed | depressed | 100.0 |
| i want to die | i want to die | 100.0 |
| edpressed | depressed | 88.9 |
| help | help | 100.0 |

Resources:
- ticket: https://praekeltorg.atlassian.net/browse/YAL-394
- sheet: https://docs.google.com/spreadsheets/d/1MpKmtDdFvc1nlF7Eu37pIKCa59DKqOR5Pr_gJcVjsyQ/edit#gid=0

# Other algorithms:
## JaroWinkler
| keyword | match |  match ratio |
|---|---|---:|
| hi | hit | 91.1 |
| depressed | depressed | 100.0 |
| i want to die | i want to die | 100.0 |
| edpressed | depressed | 96.3 |
| i am depressed | depressed | 88.1 |
| help | help | 100.0 |
| hi | hit | 91.1 |
| heita | hit | 88.0 |


KaitCrawford:
@megh-k was there any conversation with SXD about making this less disruptive? like sending an additional message with instructions to get to the counselling flow instead of sending them directly there?

megh-k:
@KaitCrawford I asked about whether some keywords weren't too broad but the conclusion was "lets see how it goes first" but I will check in again regarding an additional step

KaitCrawford:
@megh-k maybe a compromise would be that exact matches get sent there directly but fuzzy matches just get something like "are you ok? send in *help* to speak to a counsellor"

yal/utils.py
```diff
@@ -77,3 +79,25 @@ def replace_persona_fields(text, metadata={}):
...
+def check_keyword(keyword, keyword_list):
+    if process.extract(
+        keyword,
+        keyword_list,
+        scorer=fuzz.ratio,
+        score_cutoff=75,
+    ):
+        return True
+    else:
+        return False
```
rudigiesler, line 103: We can either do `return bool(process.extract(...))` or just `return process.extract(...)` to simplify this https://adamj.eu/tech/2020/01/17/simplify-your-ifs-that-return-booleans/

rudigiesler (commented):
Just one small comment, otherwise looks good :)

yal/tests/states_dictionary.md
```diff
@@ -172,6 +172,7 @@
 | state_facebook_page                        |        TRUE        |     Text     |            TRUE          | Sends user link to Facebook. User response is "menu" |
 | state_twitter_page                         |        TRUE        |     Text     |            TRUE          | Sends user link to Twitter. User response is "menu" |
 | state_website                              |        TRUE        |     Text     |            TRUE          | Sends user link to website. User response is "menu" |
+| state_confirm_redirect_please_call_me      |        TRUE        |     Text     |            FALSE         | Confirms with a user whether they'd like to speak to a human if their response fuzzy matches an emergency keyword |
```
KaitCrawford, line 175: Since this accepts a user response it should definitely be added to the flow results app

KaitCrawford (approved):
👍 just the update required to the flow results app

yal/pleasecallme.py
```diff
@@ -1088,3 +1089,37 @@ async def state_website(self):
...
+        )
+        return WhatsAppButtonState(
+            self,
+            question=question,
+            choices=[
+                Choice("help now", "please help"),
+                Choice("mistake", "go back"),
+            ],
+            error=self._(get_generic_error()),
+            next=_next,
```
rudigiesler, line 1124: We could maybe make this a bit simpler by using the `dict` format for next?
```
next={
  "help now": "state_please_call_start",
  "mistake": self.user.metadata["emergency_keyword_previous_state",
}
```

rudigiesler (approved):
Besides Kaitlyn's comment on flow results, looks good!

yal/pleasecallme.py
megh-k: @rudigiesler I was worried that if the key was missing from metadata, it might cause chaos?

rudigiesler: Well at the moment, if it's missing from metadata, our `next` function is returning `None`, which will still cause chaos. If we think this is possible, maybe we should add a test for it?

We can also do `self.user.,metadata.get("emergency_keyword_previous_state", "fallbackstatehere")` if you want a fallback for if that metadata isn't set

PR #459, opened 2022-11-21
Add survey section 1 copy

erikh360:


yal/data/seqmentation_survey_questions.py
```diff
@@ -1,7 +1,36 @@
 SURVEY_QUESTIONS = {
     "1": {
-        "start": "state_s1_4_income",
+        "start": "state_s1_1_gender",
         "questions": {
+            "state_s1_1_gender": {
```
rudigiesler, line 5: Not for this PR, but something like gender (and other questions) we might want to save to the contact profile to use at a later stage. Have we thought about how we want handle answers that we might want to use, vs putting them in the answer store for later analysis?

rudigiesler (approved):

yal/data/seqmentation_survey_questions.py
erikh360: It was the idea at some point but the values aren't the same as what we have in our profile.

PR #463, opened 2022-11-22
State checker test and data dictionary

KaitCrawford:
This is a test that checks if all the states are present in the data dictionary doc.
The idea is to encourage engineers to ensure that any changes to the list of states are documented and added to the flow results app
(see the latest failed test for an example of what it will look like)
Future work for this is to use the doc to generate the flow results specification to import into the flow results app


KaitCrawford:
extract from the xlsx file for example
image

yal/tests/test_main.py
```diff
@@ -73,6 +74,60 @@ def test_no_state_name_clashes():
...
+        | cp_states
+        | q_states
+        | pc_states
+        | sf_states
+        | aaq_states
+        | fb_states
+        | c_fb_states
+        | sf_s_states
+        | wa_fb_states
+    )
```
rudigiesler, line 116: Should we maybe not extract this into a function? Otherwise there are two places to update this list, which makes it easy to miss.

yal/tests/states_dictionary.md
```diff
@@ -0,0 +1,5 @@
+| state_name                | accepts_user_input | added_to_flow_results_app | description                                                                       |
+|---------------------------|--------------------|---------------------------|-----------------------------------------------------------------------------------|
+| state_display_preferences |        TRUE        |            TRUE           | displays profile fields to user. User response is string of which field to update |
+| state_update_age          |        TRUE        |            TRUE           | asks user to enter their age. User response is int                                |
+| state_update_age_confirm  |        TRUE        |            TRUE           | asks user to confirm new age choice. User response is string "yes" or "no"        |
```
rudigiesler, line 5: I think this is sufficient to generate the flow results specification, as at the moment, we've got all the types set to `text`. So all we need is the identifier, `state_name`; `type` hardcoded as `text` which has no additional options; and the label, which we can set to the `description` here.

Ideally though, we would be able to specify the types as well, but doing that in a table might be a bit difficult.

eg. for `state_update_age_confirm`, a better type would be `select_one`, which has an additional required field `choices`, which we would want to set to `["yes","no"]`. 

For `state_update_age`, there's the `numeric` type, which has an optional field `range`, which we might want to set to `[1,100]`. Basically codifying what we have in the `description` field in this table.

https://floip.gitbook.io/flow-results-specification/specification#select_one

PR #465, opened 2022-11-24
Get RapidPro contact on inbound message

rudigiesler:
Previously we fetched the RapidPro contact when we thought we might need it, by keeping a timestamp locally, and checking when we got close to the timestamp. This is complicated, and won't work for some future changes that we want to do, so this PR changes to fetching the RapidPro contact fields on every inbound message, and storing them on the user metadata

### DIFF
--- a/yal/tests/test_askaquestion.py
+++ b/yal/tests/test_askaquestion.py
@@ -94,7 +94,7 @@ async def aaq_mock():
         config.AAQ_URL = url
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
 async def rapidpro_mock():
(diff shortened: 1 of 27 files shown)


yal/tests/test_askaquestion.py
```diff
@@ -94,7 +94,7 @@ def add_feedback(request):
         config.AAQ_URL = url
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
```
erikh360, line 97: Nice!

erikh360 (approved):
Cool, this looks much cleaner

PR #466, opened 2022-11-24
Adding privacy policy to main menu for first time users

megh-k:


yal/mainmenu.py
```diff
@@ -58,6 +58,15 @@ async def get_suggested_choices(self, parent_topic_links={}):
...
         return suggested_choices
 
+    async def get_first_time_user(self, whatsapp_id):
+        first_time_user = self.user.metadata.get("first_time_on_main_menu")
+        error = await rapidpro.update_profile(
+            whatsapp_id, {"first_time_on_main_menu": ""}, self.user.metadata
+        )
+        if error:
+            return await self.go_to_state("state_error")
+        return bool(first_time_user)
```
rudigiesler, line 68: What if we do this the other way around, then it will also work for existing users? eg. if `seen_privacy_policy` is blank, then we send the extra message, and set `seen_privacy_polity` to `TRUE`, so that we won't send it again.

rudigiesler (commented):
Looks good! I have a little suggestion for making it work for first time users. I'm not sure if it would work, or if there's something I'm not considering. Let me know what you think

yal/mainmenu.py
```diff
@@ -58,6 +58,15 @@ async def get_suggested_choices(self, parent_topic_links={}):
 
         return suggested_choices
 
+    async def get_privacy_reminder_sent(self, whatsapp_id):
+        privacy_reminder_sent = self.user.metadata.get("privacy_reminder_sent")
+        error = await rapidpro.update_profile(
+            whatsapp_id, {"privacy_reminder_sent": "True"}, self.user.metadata
+        )
```
rudigiesler, line 65: I think this means that we'll be updating the profile every time we call this method, ie. even if it's already `True`, we'll still make the call to update it. Maybe we can put this request in a `if not privacy_reminder_sent:`

rudigiesler (commented):
Just one small comment on reducing the amount of calls to RapidPro, but otherwise looks good :)

yal/tests/test_onboarding.py
```diff
@@ -429,3 +429,20 @@ async def test_onboarding_reminder_later_response_actioned(
             "last_onboarding_time": "2022-06-19T17:30:00",
         },
     }
+
+
+@pytest.mark.asyncio
+async def test_onboarding_set_first_time_menu(
```
KaitCrawford, line 435: I don't think the name for this test is correct?

KaitCrawford (approved):
small comment but 👍 when you've made Rudi's change

rudigiesler (approved):

PR #470, opened 2022-12-01
Fix None metadata

rudigiesler:
This started off as a fix for the one error I was seeing, which was that metadata fields could be None, and so you can't call lower on them for the replace_persona_fields function.

It turned into a bit of a rework, firstly to make testing a bit easier in test_main, and secondly to only save a rapidpro contact field to metadata if it has a value (which should also help if we're making this metadata can't be none assumption anywhere else)

### DIFF
--- a/yal/main.py
+++ b/yal/main.py
@@ -73,7 +73,10 @@ class Application(
         if error:
             return await self.go_to_state("state_error")
         for key, value in fields.items():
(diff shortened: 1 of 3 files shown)


megh-k (approved):

KaitCrawford (commented):
👍 late approval from me as well

PR #471, opened 2022-12-01
fix privacy reminder already sent

rudigiesler:
There's a "variable referenced before assignment" error if the user has already seen the privacy statement, caused by an indentation error

This PR adds a test that fails with the old code, fixes the indentation, and does some refactoring with the rapidpro mock to make tests easier to understand


KaitCrawford (approved):
👍 looks good to me

PR #473, opened 2022-12-07
YAL: add onboarding and assessments content

megh-k:
Notes: 
- the new onboarding is simplified and "gender_other" is removed 
- relationship status has been added to onboarding 
- onboarding goes straight to the first assessment, onboarding is submitted just before it goes to the assessments 

Tasks:
- [x] Add new question types for whatsapp list and button messages
- [x] Use the same generic assessment flow for multiple assessments
- [ ] Add triggers + surrounding flows for each of the assessments (they're not just assessments, there's things surrounding the assessment that need to be hardcoded in)
- [ ]  Add scoring
- [x] states for opt in for push messages 
- [x] remove tests for assessments that contain content
- [ ] go back over miroboard to ensure all changes to content are caught
- [ ] add all requested contact fields from DS 
- [ ] reminders

### DIFF
--- a/yal/pushmessages_optin.py
+++ b/yal/pushmessages_optin.py
@@ -49,7 +49,7 @@ class Application(BaseApplication):
         msisdn = normalise_phonenumber(self.inbound.from_addr)
         whatsapp_id = msisdn.lstrip(" + ")
         data = {
-            "opted_in": "False",
+            "push_message_opt_in": "False",
         }
         error = await rapidpro.update_profile(whatsapp_id, data, self.user.metadata)
         if error:
@@ -75,7 +75,7 @@ class Application(BaseApplication):
         msisdn = normalise_phonenumber(self.inbound.from_addr)
         whatsapp_id = msisdn.lstrip(" + ")
         data = {
-            "opted_in": "True",
+            "push_message_opt_in": "True",
         }
         error = await rapidpro.update_profile(whatsapp_id, data, self.user.metadata)
         if error:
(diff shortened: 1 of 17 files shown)


yal/pushmessages_optin.py
```diff
@@ -0,0 +1,122 @@
...
+                "yes": "state_pushmessage_optin_yes_submit",
+                "no": "state_pushmessage_optin_no_submit",
+            },
+        )
+
+    async def state_pushmessage_optin_no_submit(self):
+        msisdn = normalise_phonenumber(self.inbound.from_addr)
+        whatsapp_id = msisdn.lstrip(" + ")
+        data = {
+            "opted_in": "False",
```
megh-k, line 52: @rudigiesler we have a "Opt out" flow, which resets all the notification timers, and now we've implemented this, I have no idea if these two things are meant to have the same functionality and if push messages refers to the scheduled notifications or just the surveys, etc, and what that entails for the service. Please advise on whether there is a better naming convention that I can use in this section

rudigiesler: Yeah maybe we can be more explicit about what it is for? `segmentation_opt_in` or `push_message_opt_in` or something like that? That way we're explicit that this is only for the new push message things, and not for any of the notifications/reengagement messages.

I think we probably want users to be able to opt out of one type but keep receiving the other. But also a place where you can opt out of everything?

megh-k: Ok cool :) I will update the opt out flow to include these :)

megh-k (commented):
👍

rudigiesler (approved):

rudigiesler:
Just to summarise here what was discussed. We got this feature to a point where it triggers a working assessment, but didn't want this PR to get too big, so we're splitting out this feature into multiple PRs

PR #475, opened 2022-12-08
Yal update assessment content

megh-k:


yal/assessment_data/body_image.py
```diff
@@ -0,0 +1,53 @@
+# Assessment state naming convention
+# state_a[number]_[subsection]_q[question]_topic
+ASSESSMENT_QUESTIONS = {
+    "8":
```
rudigiesler, line 4: Shouldn't this be 1?

yal/assessment_data/body_image.py
```diff
@@ -0,0 +1,53 @@
...
+                    [
+                        "*_I feel good about my body._*",
+                    ]
+                ),
+                "options": [
+                    "Strongly Agree",
+                    "Kinda Agree",
+                    "Do not Agree",
+                ],
+                "next": None,
```
rudigiesler, line 49: We probably need `type`s for these questions, otherwise they get treated like `info`, so they'll only show the text and then not wait for user input

yal/assessment_data/locus_of_control.py
```diff
@@ -0,0 +1,295 @@
...
+                    [
+                        "🏁 🎉",
+                        "",
+                        "[persona_emoji] Yoh! That was hella questions and a "
+                        "looot of personal and sensitive stuff too, eh?",
+                        "",
+                        "Thanks so much — proud of you for pushing through. 💪",
+                    ]
+                ),
+                "next": "state_a2_8_selfesteem_2_worth",
```
rudigiesler, line 291: Shouldn't this be `None`? Since it's the end of the assessment

rudigiesler (commented):
Just a few comments. I think most of the question types should be `list` or `button`? At least looking at the Miro board, only the sexual health literacy assessment uses `choice`

yal/assessment_data/connectedness.py
```diff
@@ -0,0 +1,27 @@
...
+                    [
+                        "[persona_emoji] *Do you have someone to talk to when you "
+                        "have a worry or problem?*",
+                    ]
+                ),
+                "options": [
+                    "Never",
+                    "Sometimes",
+                    "Most of the time",
+                    "All the time",
```
rudigiesler, line 21: We can only have 3 buttons per message, so I think this needs to be `list` type question (and the same for any other button questions with more than 3 options)

rudigiesler (commented):
Just one comment about some other the button messages, otherwise looks good :)

yal/assessment_data/A2_locus_of_control.py
```diff
@@ -0,0 +1,110 @@
...
+                    "Very true",
+                    "Extremely True",
+                ],
+                "scoring": {
+                    "not_at_all_true": 0,
+                    "a_little_true": 0,
+                    "kind_of_true": 0,
+                    "very_true": 3,
+                    "extremely_true": 5,
+                },
```
rudigiesler, line 28: I see we have scoring for one question here, but none of the others. Should we maybe just use the same scoring here for all of the questions?

PR #481, opened 2022-12-14
Finish assessment later

megh-k:


yal/assessments.py
```diff
@@ -150,3 +154,42 @@ async def state_survey_process_answer(self):
...
+                ]
+            )
+        )
+        return WhatsAppButtonState(
+            self,
+            question=question,
+            choices=[Choice("menu", "Go to main menu")],
+            error=get_generic_error(),
+            next=MainMenuApplication.START_STATE,
+        )
```
rudigiesler, line 195: I'm not sure about these states existing in `assessments`. They don't seem to interact with the rest of the generic assessment states. I think `onboarding` is probably a better place for this? Since it's specific to that flow?

yal/assessments.py
```diff
@@ -150,3 +154,42 @@ async def state_survey_process_answer(self):
                 self.save_metadata("assessment_score", score)
 
         return await self.go_to_state("state_survey_question")
+
+    async def state_assessment_later_submit(self):
+        msisdn = normalise_phonenumber(self.inbound.from_addr)
+        whatsapp_id = msisdn.lstrip(" + ")
+
+        reminder_time = get_current_datetime() + timedelta(hours=23)
+        self.save_metadata("assessment_reminder", reminder_time.isoformat())
```
rudigiesler, line 163: The `update_profile` will also update this metadata, so we don't need to worry about doing that here.

yal/assessments.py
```diff
@@ -150,3 +154,42 @@ async def state_survey_process_answer(self):
...
+        assessment_name = self.user.metadata.get(
+            "assessment_name", "sexual_health_literacy"
+        )
+
+        data = {
+            "assessment_reminder": reminder_time.isoformat(),
+            "assessment_name": assessment_name,
```
rudigiesler, line 170: Could we maybe add these to `docs/contact_fields.md` so long, so that we can start keeping track of all the rapidpro contact fields and how they are used.

rudigiesler (commented):
Just one comment, otherwise looks good

yal/assessments.py
megh-k: Oh! don't we do that for other things? I copy pasta'd from another flow :)

yal/assessments.py
megh-k: I do agree, but, all the flows follow the same "I'll do this later" flow where they set a reminder and ask the user if they'd like to return to the main menu 
https://miro.com/app/board/uXjVP-lrlK0=/?moveToWidget=3458764541062653134&cot=14

yal/assessments.py
rudigiesler: Ah, I must've missed one then. Yeah I combined it so that the metadata and the rapidpro contact wouldn't be able to go out of sync.

yal/onboarding.py
```diff
@@ -287,9 +290,80 @@ async def state_sexual_literacy_assessment_end(self):
...
+            risk = "high_risk"
+        else:
+            # score of 25-30 low risk
+            risk = "low_risk"
+
+        msisdn = normalise_phonenumber(self.inbound.from_addr)
+        whatsapp_id = msisdn.lstrip(" + ")
+        data = {
+            "sexual_health_lit_risk": risk,
+        }
```
rudigiesler, line 308: I'm wondering if we should be storing the score on the contact field? Otherwise right now it's lost at the end of this session. That's what I've got for the default behaviour of the other assessments, storing just the score on the contact field.

rudigiesler (commented):
Just one small comment, otherwise looks good :)

PR #487, opened 2022-12-19
Add payload triggers

rudigiesler:
- Add payload routing
- Add fork to github branch build list

### DIFF
--- a/yal/main.py
+++ b/yal/main.py
@@ -128,7 +128,18 @@ class Application(
             message.session_event = Message.SESSION_EVENT.RESUME
             self.state_name = feedback_state
 
+        # Replies to template push messages
+        payload = utils.get_by_path(
+            message.transport_metadata, "message", "button", "payload"
+        )
+        if payload and payload.startswith("state_") and payload in dir(self):
+            self.user.session_id = None
+            self.state_name = payload
+
         return await super().process_message(message)
+
+    async def state_testing_payload(self):
+        return EndState(self, "Payload received")
 
     async def state_qa_reset_feedback_timestamp_keywords(self):
         self.save_metadata("feedback_timestamp", get_current_datetime().isoformat())
--- a/yal/main.py
+++ b/yal/main.py
@@ -139,8 +139,35 @@ class Application(
 
         return await super().process_message(message)
 
-    async def state_testing_payload(self):
-        return EndState(self, "Payload received")
+    async def state_sexual_health_literacy_assessment(self):
+        return EndState(
+            self, "Payload received: state_sexual_health_literacy_assessment"
+        )
+
+    async def state_locus_of_control_assessment(self):
+        return EndState(self, "Payload received: state_locus_of_control_assessment")
+
+    async def state_depression_and_anxiety_assessment(self):
+        return EndState(
+            self, "Payload received: state_depression_and_anxiety_assessment"
+        )
+
+    async def state_connectedness_assessment(self):
+        return EndState(self, "Payload received: state_connectedness_assessment")
+
+    async def state_gender_attitude_assessment(self):
+        return EndState(self, "Payload received: state_gender_attitude_assessment")
+
+    async def state_body_image_assessment(self):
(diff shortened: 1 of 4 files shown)


yal/main.py
```diff
@@ -128,8 +128,19 @@ async def process_message(self, message):
             message.session_event = Message.SESSION_EVENT.RESUME
             self.state_name = feedback_state
 
+        # Replies to template push messages
+        payload = utils.get_by_path(
+            message.transport_metadata, "message", "button", "payload"
+        )
+        if payload and payload.startswith("state_") and payload in dir(self):
```
KaitCrawford, line 135: I think we need to remove the `and payload in dir(self)` from here because some of the states won't exist in the codebase yet.

yal/main.py
```diff
@@ -128,8 +128,19 @@ async def process_message(self, message):
...
+            self.user.session_id = None
+            self.state_name = payload
```
KaitCrawford, line 137: Shouldn't this direct the user to `state_testing_payload`?

KaitCrawford:
@rudigiesler just a few additional thoughts I had while working on this:
- can we wrap the code in main with a feature or user flag eg `if self.user.metadata.get("is_tester", "false") == "true":`?
- can we move the code in main higher up so that the payload takes precedence over whatever content is in the button?
- can the response message have the payload content in it so that it's easier to verify that things have gone well?

yal/main.py
rudigiesler: I was thinking that for the production test, we make the payload `state_testing_payload`, since all the other states that it would normally direct to aren't in this deploy. I also wanted to be able to test the actual routing functionality in production, not just hardcode it to always go to the same state.

yal/main.py
rudigiesler: That's probably what we want though isn't it? We don't want to change what we're doing if it will result in us sending the user an error, we rather stick with existing states for testing (which is why I added the test state, so that we have something to test with). I also wanted to try and make this test as close to the actual code as possible

yal/main.py
KaitCrawford: @rudigiesler but the intention of this is to allow us to test the template messages? I'd prefer to ensure that the states are being added to the payload correctly. And that different buttons send different payloads.
That also means that I can add the correct payloads to the templates now and I don't have to go through the flows to update the payloads when we need to deploy

rudigiesler:
@KaitCrawford Yup, we can wrap it up in a feature flag, I'll do that now
For moving it up, that will actually make it have lower precedence, we want it right at the bottom (and that's what the actual code will be like, so I want to keep it as similar as possible)
For the payload content, that will only make sense if we redirect all payload messages to the same state. I'm worried that will be changing the feature too much, and so we'll run into issues if our testing differs too far.

yal/main.py
rudigiesler: Ah I see what you're saying. My intention with this was to be able to test this bit of code, that we receive the payload correctly from whatsapp, and that the routing to a state code works as expected, I wasn't thinking of testing the actual template sending webhooks.

What about if we do both? If I add the same states that we have in the main branch, but here they're just an endstate, instead of being an assessment?

yal/main.py
```diff
@@ -128,8 +128,47 @@ async def process_message(self, message):
             message.session_event = Message.SESSION_EVENT.RESUME
             self.state_name = feedback_state
 
+        if self.user.metadata.get("is_tester") == "TRUE":
```
KaitCrawford, line 131: should there be a default here so that if "is_tester" isn't in the metadata this doesn't throw an error?

KaitCrawford (commented):
Thanks @rudigiesler just one comment

yal/main.py
rudigiesler: Yeah the default here is `None`
```
>>> metadata = {}
>>> print(metadata.get("is_tester"))
None
```

KaitCrawford (approved):
👍

PR #492, opened 2022-12-20
Yal 552 extra message after onboarding

megh-k:


yal/onboarding.py
```diff
@@ -252,7 +252,9 @@ async def state_sexual_literacy_assessment_start(self):
             )
         )
         await self.publish_message(msg)
-        await asyncio.sleep(0.5)
```
rudigiesler, line 255: Do we maybe want to keep this sleep? I think that's to ensure that the messages don't arrive out of order

rudigiesler (approved):

PR #499, opened 2022-12-22
Rename assessment_name contact field to assessment_reminder_name

rudigiesler:
Having it be assessment_name overwrites the metadata field that we're using to keep state in the assessments, so we can't use it as a contact field name, so I've renamed it to fix that


KaitCrawford (approved):
👍

PR #500, opened 2022-12-22
Swap the locus of control and sexual health literacy assessments

megh-k:


yal/assessments.py
```diff
@@ -65,7 +65,7 @@ async def state_survey_question(self):
 
         section = str(metadata.get("assessment_section", "1"))
 
-        assessment_name = metadata.get("assessment_name", "sexual_health_literacy")
+        assessment_name = metadata.get("assessment_name", "locus_of_control")
```
KaitCrawford, line 68: This seems like a generic assessment application, why are we setting specific assessment names here?

KaitCrawford (approved):
👍 you can merge but could you just answer my question so that I understand that piece 🙏

yal/assessments.py
megh-k: @KaitCrawford this is because its the first assessment, @rudigiesler can probably give more specifics?

rudigiesler: It was because this used to be for just one assessment, `sexual_health_literacy`. I didn't have time to go through and find everywhere where it was being called, and set the `assessment_name`, so I had it set to a default assessment if no specific assessment was set.

PR #510, opened 2023-01-05
Send the user to a generic state that asks them what they would like …

megh-k:
…to do next after the assessments

### DIFF
--- a/yal/main.py
+++ b/yal/main.py
@@ -2,8 +2,8 @@ import asyncio
 import logging
 
 from vaccine.models import Message
-from vaccine.states import EndState
-from vaccine.utils import random_id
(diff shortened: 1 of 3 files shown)


KaitCrawford (commented):
Sorry to be a pain but we should really have tests for every assessment end state going to the generic message.
Otherwise it looks good

KaitCrawford (commented):
Sorry to be pedantic but please add tests for the "ask a question" and "update settings" buttons as well

KaitCrawford (approved):
👍 looks good.
but I'd ask that you pls don't land this PR until my one has passed QA 🙏 Mine must urgently be released today so we need to keep the main branch clear in case there are issues

PR #522, opened 2023-01-24
Fix loop in please call me by removing emergency keyword that creates a loop

megh-k:


KaitCrawford (approved):
👍 Did you check this with whoever drew up the list of emergency keywords?

megh-k:
@KaitCrawford I checked with Chuk and he said its not a problem :)

PR #525, opened 2023-01-30
Add validation to persona emoji in onboarding

megh-k:


yal/onboarding.py
```diff
@@ -92,8 +94,15 @@ async def state_persona_emoji(self):
         )
 
     async def state_save_persona_emoji(self):
+        def extract_first_emoji(persona_emoji):
+            emojs_present = "".join(c for c in persona_emoji if c in EMOJI_DATA)
+            if emojs_present:
+                return emojs_present[0]
+            return ""
```
rudigiesler, line 101: Extracting only the first character has unintended results for emoji's with modifiers:

```
>>> extract_first_emoji("👍🏿")
'👍'
```

So we might need to take a different approach with this, as well as adding some unit tests for these use cases.

I think we can use the `emoji_list` function from the `emoji` library, which should do better:
```
>>> from emoji import emoji_list
>>> emoji_list("👍🏿foo😀😀")
[{'match_start': 0, 'match_end': 2, 'emoji': '👍🏿'}, {'match_start': 5, 'match_end': 6, 'emoji': '😀'}, {'match_start': 6, 'match_end': 7, 'emoji': '😀'}]
```

megh-k: @rudigiesler yeah I ran into this issue this morning :) and this was my fix! I didnt realise you had reviewed the code 😂

yal/utils.py
```diff
@@ -76,11 +77,22 @@ def normalise_phonenumber(phonenumber):
         raise ValueError("Invalid phone number")
 
 
+def extract_first_emoji(persona_emoji):
+    if is_emoji(persona_emoji):
+        return persona_emoji
```
rudigiesler, line 82: Do we still need this `is_emoji` check? Could we not just rely on the `emoji_list` function completely?

yal/change_preferences.py
```diff
@@ -633,8 +638,7 @@ async def state_update_bot_emoji_submit(self):
         choice = self.user.answers.get("state_update_bot_emoji")
         if choice == "skip":
             return await self.go_to_state("state_display_preferences")
-
-        data = {"persona_emoji": choice}
+        data = {"persona_emoji": extract_first_emoji(choice)}
```
rudigiesler, line 641: I don't think we were doing this before, but should we handle if the user doesn't send any emoji's? I guess we'll just use the default in that case, so maybe it doesn't matter?

yal/utils.py
```diff
@@ -76,11 +77,22 @@ def normalise_phonenumber(phonenumber):
...
+
+
 def replace_persona_fields(text, metadata={}):
     for key in PERSONA_FIELDS:
         value = metadata.get(key)
         if value and value.lower() != "skip":
-            text = text.replace(f"[{key}]", value)
+            if key == "persona_emoji":
+                value = extract_first_emoji(value)
+            text = text.replace(f"[{key}]", "".join(value.splitlines()).strip())
```
rudigiesler, line 95: Are we trying to replace newlines with spaces using `splitlines`? What about `re.sub(r"\s+", " ", value)`, which will also replace any multiple spaces with a single space, and multiple newlines with a single space?

rudigiesler (approved):

PR #531, opened 2023-02-01
Add script to update the rapidpro contacts persona_emoji

megh-k:


yal/scripts/update_rapidpro_contacts_persona_emoji.py
```diff
@@ -0,0 +1,63 @@
...
+import emoji
+import requests
+
+from yal.utils import extract_first_emoji
+
+
+def get_arguments() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Pretends to be a webhook from whatsapp for a template message "
+        "button click with payload"
```
rudigiesler, line 13: Should probably change this description to something that matches what this does.

yal/scripts/update_rapidpro_contacts_persona_emoji.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    parser.add_argument("url", help="Base URL for the transport")
```
rudigiesler, line 15: for the RapidPro API?

yal/scripts/update_rapidpro_contacts_persona_emoji.py
```diff
@@ -0,0 +1,63 @@
...
+    rapidpro_base_api_url = arguments.url
+    api_token = arguments.api_token
+
+    response = requests.get(
+        f"{rapidpro_base_api_url}/api/v2/contacts.json",
+        headers={"Authorization": f"Token {api_token}"},
+    )
+    response = response.json()
+
+    for contact in response["results"]:
```
rudigiesler, line 31: This doesn't seem to handle pagination. I think we'll have to go through all the pages to get all the contacts. The way I like doing pagination is something like:
```python
list_contacts_url = f"{rapidpro_base_api_url}/api/v2/contacts.json"

while list_contacts_url:
   response = requests.get(list_contacts_url, ...).json()
   for contact is response["results"]:
       ...
   list_contacts_url = response["next"]
```

yal/scripts/update_rapidpro_contacts_persona_emoji.py
```diff
@@ -0,0 +1,63 @@
...
+        if (
+            not emoji.is_emoji(contact["fields"]["persona_emoji"])
+            and contact["fields"]["persona_emoji"] != "Skip"
+            and contact["fields"]["persona_emoji"] is not None
+        ):
+            new_persona_emoji = extract_first_emoji(contact["fields"]["persona_emoji"])
```
rudigiesler, line 37: We should probably check if `persona_emoji == new_persona_emoji`, and if so, we can skip the request to update the contact, to avoid updating the field unnecessarily (especially since we think this applies to a small amount of contacts)

yal/scripts/update_rapidpro_contacts_persona_emoji.py
```diff
@@ -0,0 +1,63 @@
...
+    response = requests.get(
+        f"{rapidpro_base_api_url}/api/v2/contacts.json",
+        headers={"Authorization": f"Token {api_token}"},
+    )
+    response = response.json()
+
+    for contact in response["results"]:
+        if (
+            not emoji.is_emoji(contact["fields"]["persona_emoji"])
+            and contact["fields"]["persona_emoji"] != "Skip"
```
rudigiesler, line 34: Should this maybe be a case insensitive check? So a:
```python
not emoji.is_emoji(contact["fields"]["persona_emoji"])
and contact["fields"]["persona_emoji"]
and contact["fields"]["persona_emoji"].lower() != "skip"
```

yal/scripts/update_rapidpro_contacts_persona_emoji.py
rudigiesler: Ah right, yeah I can't think of any values where it would make it into the if and not be changed

rudigiesler (approved):

PR #532, opened 2023-02-08
Suggested content api calls are logging errors in sentry when we make…

megh-k:
… invalid requests

The api call wont be made when the topics_viewed is empty
The error wont be logged as an exception, only as a warning, when the page_id is that of a content page index. This needs to still be updated in contentrepo for better error handling

there is one other place where this takes place, but this needs further discussion

### DIFF
--- a/yal/contentrepo.py
+++ b/yal/contentrepo.py
@@ -85,7 +85,7 @@ async def get_choices_by_path(path: str) -> Tuple[bool, List[Choice]]:
                 break
             except HTTP_EXCEPTIONS as e:
                 if i == 2:
-                    logger.exception(e)
+                    logger.warning(e)
                     return True, []
                 else:
                     continue
--- a/yal/contentrepo.py
+++ b/yal/contentrepo.py
@@ -84,6 +84,8 @@ async def get_choices_by_path(path: str) -> Tuple[bool, List[Choice]]:
 
                 break
             except HTTP_EXCEPTIONS as e:
+                # TODO: better error handling once contentrepo is updated to
+                # return 404 errors on page not found
                 if i == 2:
                     logger.warning(e)
                     return True, []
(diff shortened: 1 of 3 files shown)


yal/contentrepo.py
```diff
@@ -85,7 +85,7 @@ async def get_choices_by_path(path: str) -> Tuple[bool, List[Choice]]:
                 break
             except HTTP_EXCEPTIONS as e:
                 if i == 2:
-                    logger.exception(e)
+                    logger.warning(e)
```
rudigiesler, line 88: This seems a bit dangerous. Now if there are any other errors, we won't know about them.

The case for this is that if we get a `404`, we want to return an empty list right? So we could handle that case separately. Check if the response is a 404, and if it is, then return an empty list (I'm not sure what makes sense for the first argument here, will have to look at where it's being used), otherwise we log the exception as normal, since it's not something we're expecting.

That might also make the change in the other file unnecessary, since we'll be returning an empty list.

megh-k: @rudigiesler 100% agree but for now, because we can't update contentrepo yet to throw 404 errors when a page isnt found, we wont be able to look for 404's when a page isnt found, this will have to be a change made later down the line because any update of contentrepo is blocked by SXD not giving me any feedback on the import export feature

rudigiesler: @megh-k Hmm okay, and will the change in the other file not stop contentrepo giving us an error in the first place, making this unnecessary? If not, then can we just add a TODO comment here explaining to change it back once we've fixed contentrepo?

megh-k: @rudigiesler I think only if topics viewed is empty? this is more to handle when we request a content page index instead of a content page (the overwhelming number of errors is coming from 1019 which is a content index page and thus isnt found in the content page query on the contentrepo side)

rudigiesler: @megh-k Ah I see. 👍 if you just add a TODO comment so that this doesn't get forgotten/lost.

rudigiesler (approved):

PR #535, opened 2023-02-16
Fix content titles for related pages

rudigiesler:
Before, we used the titles returned from ContentRepo in the
"related_pages" key. The issue with this is because we're requesting
"whatsapp=true", we're getting the whatsapp titles, not the content
titles.

In this case we want the content titles instead, since the WhatsApp
titles include breadcrumbs, and those are too long for interactive
buttons on WhatsApp.


KaitCrawford (approved):
👍

PR #536, opened 2023-02-20
Add later state for each assessment that makes the assessment names a…

megh-k:
…vailable

### DIFF
--- a/yal/assessments.py
+++ b/yal/assessments.py
@@ -207,6 +207,14 @@ class Application(BaseApplication):
             self.save_metadata("assessment_section", section + 1)
             self.save_metadata("assessment_question_nr", 1)
             self.delete_metadata("assessment_question")
+            # clear assessment reminder info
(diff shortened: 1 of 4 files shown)


yal/assessments.py
```diff
@@ -207,6 +208,14 @@ async def state_survey_process_answer(self):
             self.save_metadata("assessment_section", section + 1)
             self.save_metadata("assessment_question_nr", 1)
             self.delete_metadata("assessment_question")
+            # clear assessment reminder info
+            data = {
```
KaitCrawford, line 212: Is it possible that these will be empty when a user gets here? Maybe there should be a check (eg against what is in the metadata) and you only do the update if they aren't empty

KaitCrawford (commented):
One comment to check we're not updating fields that are already empty

KaitCrawford (approved):
👍

PR #539, opened 2023-02-22
Update optout fields

fritzbrand:
At present, when a user chooses to delete their data during opt-out, we clear the values from a set list of hardcoded fields.

What we want to do, is pull the list of all fields from the RapidPro instance, and then clear all of them, EXCEPT a specified list of fields set to retain values


yal/optout.py
```diff
@@ -177,33 +177,27 @@ async def state_delete_saved(self):
         msisdn = utils.normalise_phonenumber(self.inbound.from_addr)
         whatsapp_id = msisdn.lstrip(" + ")
 
-        data = {
-            "onboarding_completed": "",
-            "opted_out": "TRUE",
+        rp_fields = await rapidpro.get_instance_fields()
+        rp_field_keys = [rp_field["key"] for rp_field in rp_fields]
+        rp_field_keys_set = set(rp_field_keys)
+        print("I A BEING CALLED")
```
erikh360, line 183: Probably don't need this

.github/workflows/build.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - update-optout-fields
```
erikh360, line 7: I assume this was temporarily needed?

yal/rapidpro.py
```diff
@@ -105,3 +105,28 @@ async def start_flow(whatsapp_id, flow_uuid):
...
+                if len(response_body["results"]) > 0:
+                    fields = response_body["results"]
+                break
+            except HTTP_EXCEPTIONS as e:
+                if i == 2:
+                    logger.exception(e)
+                    return True
+                else:
+                    continue
+    return fields
```
erikh360, line 132: I would instantiate `fields` before the async block above, because there is a possibility that we will reach this line without it being created.

yal/tests/test_onboarding.py
```diff
@@ -50,6 +50,18 @@ def update_contact(request):
         tstate.requests.append(request)
         return response.json({}, status=200)
 
+    @app.route("/api/v2/fields.json", methods=["GET"])
+    def get_fields(request):
+        return response.json(
+            {
```
erikh360, line 56: Is this response body correct? It's different to the fields.json GET [here](https://github.com/praekeltfoundation/vaccine-eligibility/pull/539/files#diff-65eaca890d8cecebf033b77438466aee345340a1ddbb0d0a6a7857bef8b5f3f0R97)

yal/optout.py
```diff
@@ -177,33 +177,27 @@ async def state_delete_saved(self):
         msisdn = utils.normalise_phonenumber(self.inbound.from_addr)
         whatsapp_id = msisdn.lstrip(" + ")
 
-        data = {
-            "onboarding_completed": "",
-            "opted_out": "TRUE",
+        rp_fields = await rapidpro.get_instance_fields()
+        rp_field_keys = [rp_field["key"] for rp_field in rp_fields]
+        rp_field_keys_set = set(rp_field_keys)
```
erikh360, line 182: I like the python list comprehension here.

You could do this in one line:
`rp_field_keys = set([rp_field["key"] for rp_field in rp_fields])`

Then just use `rp_field_keys` going forward

fritzbrand: 👍

yal/optout.py
fritzbrand: 👍

yal/tests/test_onboarding.py
fritzbrand: I removed this part, I believe it was a previous attempt of mine and is not being used at present.  Will double check during my dev testing once I have this on QA

.github/workflows/build.yaml
fritzbrand: Yea, I wanted to push my branch to QA so I can test, without affecting the main branch while people were busy trying to get V3 live, but I think that ship has sailed now

PR #543, opened 2023-03-08
YAL: allow payloads to contain page ids to direct the user to

KaitCrawford:


yal/main.py
```diff
@@ -180,9 +180,14 @@ async def process_message(self, message):
...
             message.transport_metadata, "message", "button", "payload"
         )
-        if payload and payload.startswith("state_") and payload in dir(self):
-            self.user.session_id = None
-            self.state_name = payload
+        if payload:
+            if payload.startswith("state_") and payload in dir(self):
+                self.user.session_id = None
+                self.state_name = payload
+            elif payload.startswith("page_id_") and payload.split("_")[-1].isdigit():
```
rudigiesler, line 187: This isn't user supplied data, so I'm not too worried, but I have had issues before with `isdigit`, since it returns True for numbers like ², which fail when you try to call `int` on them.

rudigiesler (approved):
Looks good! Just a warning about `isdigit`, but I think it's fine in this case.

yal/main.py
KaitCrawford: Ah ok, so what would you recommend for checking for a digit then? I'm not super keen on trying to cast it to an int and then catching the exception, since it generally needs to be a string everywhere anyway

rudigiesler: I think we can leave it as string, and just have a try/catch determine if it is an integer, eg.
```python
def is_integer(string: str) -> bool:
    try:
        int(string)
        return True
    except ValueError:
        return False
```

Otherwise, I *think* `isdecimal` will only return `True` if it's something we can convert to an `int`, but I'm not 100% sure on that one.

rudigiesler (approved):

PR #551, opened 2023-04-06
Added state for baseline start

fritzbrand:


yal/main.py
```diff
@@ -206,6 +206,13 @@ async def state_qa_reset_feedback_timestamp_keywords(self):
             "the message early",
         )
 
+    async def state_qa_start_baseline_survey(self):
+        self.save_metadata("assessment_name", "connectedness_v2")
+        self.save_metadata(
+            "assessment_end_state", "state_connectedness_assessment_v2_end"
+        )
+        return await self.go_to_state(AssessmentApplication.START_STATE)
+
```
KaitCrawford, line 215: I think it would be better to change line [176](https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/main.py#L176) to point to [state_invitation](https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/surveys/baseline.py#L17) instead. These start survey states are only in main.py for bad historical reasons.

KaitCrawford:
@fritzbrand I just realised that the previous PR didn't add the new application to the main.py file. You'll need to import the baseline survey app into main like [this] (https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/main.py#L8) and also add it to the [application definition] (https://github.com/praekeltfoundation/vaccine-eligibility/blob/main/yal/main.py#L84)

KaitCrawford (approved):
👍 looks great!

PR #560, opened 2023-04-14
make study participant field only visible if the user is in the study

Hlamallama:


fritzbrand:
@Hlamallama can we maybe change that "study_optin" variable to be a little more descriptive of the study we are talking about, something like "baseline_optin"?

KaitCrawford (commented):
@Hlamallama this looks good. Just one thing: pls add a test for if the user's study_optin field is False 🙏 that way we've got both scenario's tested

KaitCrawford:
> @Hlamallama can we maybe change that "study_optin" variable to be a little more descriptive of the study we are talking about, something like "baseline_optin"?

KaitCrawford:
> @Hlamallama can we maybe change that "study_optin" variable to be a little more descriptive of the study we are talking about, something like "baseline_optin"?

@fritzbrand I'm happy with it being more descriptive, but I don't know what works. Just want to caution against `baseline_optin` though since the study is more than the baseline assessment, so I don't think it's accurate.

fritzbrand:
@KaitCrawford yea you are right, lets leave it as is for now, so I am +1

KaitCrawford (approved):
👍 looks great!!

PR #571, opened 2023-04-18
Add method to set fields for baseline assessment reminders

KaitCrawford:
Adds a method that we can call at the start of each assessment to schedule the baseline reminders


fritzbrand:
@KaitCrawford looks good.  Can you please add a commented out example of where you would call this.  Sorry my head is down on something else, just want to keep the context of this when I get to it a bit later

KaitCrawford:
@fritzbrand I'm happy to do that. but I'm also happy to add all the calls myself in this PR, as long as you're not busy with anything in this file at the moment?

fritzbrand:
@KaitCrawford thanks, then go for it, not busy with this file at present

KaitCrawford:
Closing because new commits to the branch aren't showing up here

PR #572, opened 2023-04-18
Add method to set fields for baseline assessment reminders

KaitCrawford:
Adds a method that we can call at the start of each assessment to schedule the baseline reminders. Also adds calls to this method


fritzbrand:
👍

PR #593, opened 2023-04-20
YAL: fix i'm not interested reminder response

KaitCrawford:
Use lowercase for I'm not interested keyword and make the test more accurate so that it picks this up


fritzbrand:
👍

PR #600, opened 2023-04-24
YAL: save assessment name to flow results on start and end

KaitCrawford:
As discussed with @Themba-Gqaza :
We push the assessment name to flow results at the start and end of an assessment.


fritzbrand:
👍

PR #607, opened 2023-05-11
Baseline ethics updates

fritzbrand:


erikh360 (approved):
Looks good but we should maybe add a test for `state_monthly_household_income` state

fritzbrand:
Cool thx, will do that as part of further updates on Monday

PR #615, opened 2023-05-29
checking if the value is not empty before converting to an integer

Buhle79:


fritzbrand:
👍

erikh360:
@Buhle79 we should add a test for this please

PR #626, opened 2023-06-13
Yal 933 yal whatsapp bot prd 27 key error

Buhle79:


erikh360 (commented):
Maybe we should add a test for this? 

Not sure what would happen later on if it expects a string and gets a None

erikh360 (approved):

PR #637, opened 2023-07-05
Yal 941 yal whatsapp bot prd 2 d value error

Buhle79:


yal/utils.py
```diff
@@ -75,7 +75,7 @@ def normalise_phonenumber(phonenumber):
             assert phonenumbers.is_valid_number(pn)
             return phonenumbers.format_number(pn, phonenumbers.PhoneNumberFormat.E164)
     except (phonenumbers.phonenumberutil.NumberParseException, AssertionError):
-        raise ValueError("Invalid phone number")
+        raise ValueError("Invalid phone number {}".format(phonenumber))
```
erikh360, line 78: Can we use a f-string here?

yal/tests/test_utils.py
```diff
@@ -122,3 +123,21 @@ def test_is_integer():
 
     obj = 1234
     assert utils.is_integer(obj)
+
+
+class PhoneNumberTest(unittest.TestCase):
```
erikh360, line 128: Are we inheriting from `unittest.TestCase` to use `assertRaises`?

erikh360 (commented):
Nice, I have one comment and one question.

Also, are we just adding this to debug? It might not be a great idea to send PII to sentry?

yal/utils.py
Buhle79: Yes, I've added user phone number for debugging here.

yal/tests/test_utils.py
Buhle79: No i like to group all related test under a class

erikh360 (approved):

PR #640, opened 2023-07-19
fix endline start

Hlamallama:


yal/main.py
```diff
@@ -84,7 +84,7 @@
 QA_RESET_FEEDBACK_TIMESTAMP_KEYWORDS = {"resetfeedbacktimestampobzvmp"}
 EMERGENCY_KEYWORDS = utils.get_keywords("emergency")
 AAQ_KEYWORDS = {"ask a question"}
-EJAF_ENDLINE_SURVEY_KEYWORDS = {"answer"}
+EJAF_ENDLINE_SURVEY_KEYWORDS = {"answer", "yes  i want to answer"}
```
erikh360, line 87: Looks like there is a double space in there, can we maybe add a test for this?

fritzbrand:
👍

PR #655, opened 2023-08-08
Validation fix

Hlamallama:


yal/assessments.py
```diff
@@ -441,6 +441,15 @@ async def state_handle_assessment_reminder_response(self):
             return await self.go_to_state("state_pre_mainmenu")
 
     async def state_not_interested(self):
```
erikh360, line 443: Is this state only used in the endline survey?

erikh360 (commented):
Looks good, just the one question?

yal/assessments.py
Hlamallama: It is also used on the baseline

erikh360: Oh ok, I guess its fine to update "endline_survey_started" to "" at that point too

erikh360 (approved):

PR #679, opened 2023-08-28
check for failed whatsapp delivery

Hlamallama:


yal/main.py
```diff
@@ -147,6 +147,20 @@ async def process_message(self, message):
                 message.transport_metadata, "message", "button", "payload"
             )
 
+            whatsapp_delivery_failed = self.user.metadata.get(
```
erikh360, line 150: Can we move some of this into a separate `reset_whatsapp_delivery_failure` function

yal/main.py
```diff
@@ -147,6 +147,20 @@ async def process_message(self, message):
...
+                "whatsapp_delivery_failed", "False"
+            )
+
+            if whatsapp_delivery_failed == "True":
```
erikh360, line 154: Making this case sensitive might cause problems later on if someone updates the flow

yal/tests/test_main.py
```diff
@@ -1490,3 +1490,15 @@ async def test_phase2_payload_message_directs_to_onboarding_rel_status(
     tester.assert_state("state_rel_status")
 
     assert len(rapidpro_mock.tstate.requests) == 2
+
+
+@pytest.mark.asyncio
+async def test_whatsapp_user_reactivate(tester: AppTester, rapidpro_mock):
```
erikh360, line 1496: Maybe assert that the call to rapidpro actually happened

erikh360 (approved):

PR #684, opened 2023-09-11
fix delete saved data

Hlamallama:


erikh360 (commented):
Can we add a test for this? There might be one that we just need to check what the request looks like

erikh360 (approved):

PR #705, opened 2023-11-13
update remind me tomorrow reminders for endline

Hlamallama:


erikh360 (commented):
@Hlamallama does this update the rapidpro contact? I think it should and if we add that we should just make sure the unit test covers it

erikh360 (approved):

PR #707, opened 2023-11-14
endline reminders

Hlamallama:


erikh360 (approved):
👍 but I would prefer if we have a test for this

PR #709, opened 2023-11-14
endline reminders

Hlamallama:


yal/assessments.py
```diff
@@ -425,9 +425,11 @@ async def state_handle_assessment_reminder_response(self):
             return await self.go_to_state("state_reschedule_assessment_reminder")
 
         if inbound == "i m not interested":
-            assessment_name = ""
-            if "assessment_reminder_name" in self.user.metadata:
-                assessment_name = self.user.metadata["assessment_reminder_name"]
+            assessment_reminder_type = ""
```
erikh360, line 428: Looks like you can replace this with a .get() and a default

PR #714, opened 2023-11-22
Endline fixes

Hlamallama:


yal/main.py
```diff
@@ -233,6 +233,19 @@ async def process_message(self, message):
                 elif keyword == "i m not interested":
                     self.state_name = AssessmentApplication.NOT_INTERESTED_STATE
                 else:
+                    endline_max_limit = await rapidpro.get_rapidpro_global(
+                        "endline_max"
```
erikh360, line 237: Can we make this Global variable name a little bit more descriptive please? `Endline study max participant count` or something

yal/endline_terms_and_conditions.py
```diff
@@ -346,3 +347,30 @@ async def state_submit_terms_and_conditions_endline(self):
...
+
+        choices = [
+            Choice("menu", "Go to the menu"),
+            Choice("aaq", "Ask a question"),
+        ]
+
+        return FreeText(
+            self,
+            question=question,
+            next=choices,
```
erikh360, line 375: Doesn't look like we are testing these?

yal/rapidpro.py
```diff
@@ -187,3 +187,31 @@ async def get_global_flag(global_name):
                 else:
                     continue
     return is_active
+
+
+async def get_rapidpro_global(global_name):
```
erikh360, line 192: We're already in the rapidpro class so the name of this function can be something like `get_global_value`

erikh360 (commented):
Just a few small changes, but looks good otherwise

erikh360 (approved):

PR #725, opened 2023-12-14
Fix apostrophes

erikh360:


DevChima:
don’t or don't. I see we have the latter.
If it doesn't affect how the user sees it then.

PR #764, opened 2024-09-17
update the downvote emoji

Hlamallama:
adding a . to the down vote because turn complains about duplication ids for the buttons. looks like the emojis get striped


erikh360 (approved):
Strange that there aren't tests failing

PR #771, opened 2025-04-09
add update settings keyword

Hlamallama:


yal/main.py
```diff
@@ -265,6 +266,12 @@ async def process_message(self, message):
                 )
                 if error:
                     self.state_name = self.ERROR_STATE
+            elif (
+                keyword in UPDATE_SETTINGS
```
DevChima, line 270: What happens if  input key is "Update Settings". Are we handling cases?

DevChima: What I mean is that the UPDATE_SETTINGS keyword is "update settings" (lowercase). The keyword variable (from message.content) is cleaned using clean_inbound(). So if clean_inbound doesn't lowercase, users typing "Update Settings" or "UPDATE SETTINGS" will not trigger this state action?

DevChima (approved):

